PR #15, opened 2023-02-09
Update to Rust v1.67

tomaka:


tomaka:
CI failure caused by https://github.com/actions/checkout/issues/1169

tomaka:
I don't understand why only this PR is affected by this CI issue.

tomaka:
> I don't understand why only this PR is affected by this CI issue.

I suppose that something has changed in the Rust Docker image between 1.65 and 1.67

Issue #17, opened 2023-02-09
Figure out situation regarding resetting the runtime Wasm VM


tomaka:
Original issue: https://github.com/paritytech/smoldot/issues/760

tomaka:
See https://github.com/paritytech/substrate/blob/86c6bb9614c437b63f3dbd2afddef52f32af7866/client/executor/wasmtime/src/runtime.rs#L296 for the config that Substrate uses for wasmtime.

tomaka:
Closing after #210 and #211

I've opened other Wasm-related issues for things that are known to not match Substrate.

PR #18, opened 2023-02-09
Update to a more recent version of wasmi

tomaka:
Work in progress. Blocked by https://github.com/paritytech/wasmi/issues/658 and https://github.com/paritytech/wasmi/issues/659.

cc https://github.com/smol-dot/smoldot/issues/17


tomaka:
I have absolutely no idea why but I'm getting out of bound memory accesses errors coming from wasmi.

I have made sure that the behavior of all the memory-growth-related functions are identical to before this PR and also identical to wasmtime.

tomaka:
@Robbepop I hope you don't mind pinging you in case you have an idea, as I'm a bit clueless.

After this PR, wasmi emits a trap due to out of bound memory accesses (`Trap(Trap { reason: InstructionTrap(MemoryOutOfBounds) })`).

This didn't happen before this PR, and also doesn't happen with wasmtime.

The PR was relatively straight forward. The code in itself is also relatively straight forward.

There could be a bug in the upper layers, but then I don't see why the problem wouldn't happen with wasmtime.

It could maybe be caused because the `ext_allocator_malloc_version_1` host function is called, and the return value of that host function isn't properly injected back into the Wasm VM? Which would then make the Wasm code use an invalid pointer.

Robbepop:
Hmmm for questions like these it would have been amazing to have better debugging support built-into `wasmi`. 🤔 
I haven't yet noticed any problems with respect to out-of-bounds memory accesses from `wasmi` users such as the `contracts-pallet`.
What does the called host function `ext_allocator_malloc_version_1` exactly do and in which contexts is it called?
What instructions are causing the out of bounds errors?

tomaka:
> What does the called host function ext_allocator_malloc_version_1 exactly do and in which contexts is it called?

The Substrate/Polkadot client has its own memory allocator. It's not the Wasm code that manages its memory. Instead, it calls a host function to allocate memory and another host function to free it.
So `ext_allocator_malloc_version_1` picks a not-allocated-yet memory location and returns a pointer to it. It grows the memory if necessary.

While the behavior of this function is very complicated, my suspicion would be that the return value of that function isn't communicated properly to the Wasm code, maybe due to a bug in the new resumable functions feature. The Wasm code then uses an incorrect value as if it was the pointer.

Of course I'm basing that a suspicion on the fact that the code in smoldot has always been relatively robust, and based on what has changed between before and after this PR.

Robbepop:
Yeah it could be connected to a bug in the relatively new resumable function feature. I will probably need some more tests for it. Unfortunately there is no Wasm spec testsuite for it. :(

tomaka:
Note for myself:

The easiest way to reproduce would be to call `Metadata_metadata`. Maybe `Core_version` as well, but `Metadata_metadata` surely triggers the issue.

The next step to debug would be to log the list of all functions being called, alongside with their parameters and return value, before and after this PR, and compare.

tomaka:
My hypothesis was wrong, as the error happens after a single host function call to `ext_logging_max_level_version_1`. No call to `ext_allocator_malloc_version_1` happens before the error.

tomaka:
Disabling all Wasm features (such as `bulk-memory`) also doesn't change anything, which makes me think that the runtime doesn't any these features.

tomaka:
While trying to reproduce the issue with a minimal case, I've made it "work" (in the sense that the execution goes further)

Now to figure out what is different between the repro and the actual code.

Pasting this repro case attempt here as a self-note:

```rust
#[test]
fn repro() {
    let engine = wasmi::Engine::default();
    let module = wasmi::Module::new(
        &engine,
        &super::zstd::zstd_decode_if_necessary(
            &include_bytes!("./westend-runtime-v9300.wasm")[..],
            usize::max_value(),
        )
        .unwrap()[..],
    )
    .unwrap();
    let mut store = wasmi::Store::new(&engine, ());

    let mut linker = wasmi::Linker::<()>::new();
    linker
        .define(
            "env",
            "ext_logging_max_level_version_1",
            wasmi::Func::new(
                &mut store,
                wasmi::FuncType::new([], core::iter::once(wasmi::core::ValueType::I32)),
                |_, _, out| {
                    out[0] = wasmi::Value::I32(0);
                    Ok(())
                },
            ),
        )
        .unwrap();

    #[derive(Debug, Clone)]
    struct InterruptedTrap {
        name: String,
        parameters: Vec<wasmi::Value>,
    }
    impl core::fmt::Display for InterruptedTrap {
        fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
            write!(f, "Interrupted")
        }
    }
    impl wasmi::core::HostError for InterruptedTrap {}

    for import in module.imports() {
        match import.ty() {
            wasmi::ExternType::Func(func_type) => {
                let name = import.name().to_owned();
                let func = wasmi::Func::new(
                    &mut store,
                    func_type.clone(),
                    move |_caller, parameters, _ret| {
                        Err(wasmi::core::Trap::from(InterruptedTrap {
                            name: name.clone(),
                            parameters: parameters.to_owned(),
                        }))
                    },
                );
                let _ = linker.define(import.module(), import.name(), func);
            }
            wasmi::ExternType::Memory(memory_type) => {
                let memory = wasmi::Memory::new(&mut store, *memory_type).unwrap();
                linker
                    .define(import.module(), import.name(), memory)
                    .unwrap();
            }
            _ => unreachable!(),
        }
    }

    let instance = linker.instantiate(&mut store, &module).unwrap();
    let instance = instance.ensure_no_start(&mut store).unwrap();

    let mut out = [wasmi::Value::I64(0)];

    let mut call = instance
        .get_func(&store, "Core_version")
        .unwrap()
        .call_resumable(
            &mut store,
            &[wasmi::Value::I32(0), wasmi::Value::I32(0)],
            &mut out,
        )
        .unwrap();

    loop {
        match call {
            wasmi::ResumableCall::Resumable(r) => {
                let err = r.host_error().downcast_ref::<InterruptedTrap>().unwrap();
                if err.name == "ext_logging_max_level_version_1" {
                    call = r
                        .resume(&mut store, &[wasmi::Value::I32(0)], &mut out)
                        .unwrap()
                } else {
                    println!("{:?}", err.name);
                    todo!()
                }
            }
            wasmi::ResumableCall::Finished => todo!(),
        }
    }
}
```

tomaka:
@Robbepop 

I think I've managed to reproduce the problem.

Here's a "minimal" failing test:

```rust
#[test]
fn repro() {
    let engine = wasmi::Engine::default();
    let module = wasmi::Module::new(
        &engine,
        ... (the wasm data)...
    )
    .unwrap();
    let mut store = wasmi::Store::new(&engine, ());

    #[derive(Debug, Clone)]
    struct InterruptedTrap {
        name: String,
    }
    impl core::fmt::Display for InterruptedTrap {
        fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
            write!(f, "Interrupted")
        }
    }
    impl wasmi::core::HostError for InterruptedTrap {}

    let mut linker = wasmi::Linker::<()>::new();
    for import in module.imports() {
        match import.ty() {
            wasmi::ExternType::Func(func_type) => {
                let name = import.name().to_owned();
                let func = wasmi::Func::new(
                    &mut store,
                    func_type.clone(),
                    move |_caller, _params, _ret| {
                        Err(wasmi::core::Trap::from(InterruptedTrap {
                            name: name.clone(),
                        }))
                    },
                );
                linker.define(import.module(), import.name(), func).unwrap();
            }
            wasmi::ExternType::Memory(memory_type) => {
                let memory = wasmi::Memory::new(&mut store, *memory_type).unwrap();
                linker
                    .define(import.module(), import.name(), memory)
                    .unwrap();
            }
            _ => unreachable!(),
        }
    }

    let instance = linker.instantiate(&mut store, &module).unwrap();
    let instance = instance.ensure_no_start(&mut store).unwrap();

    let mut out = [wasmi::Value::I64(0)];

    let mut call = instance
        .get_func(&store, "Core_version")
        .unwrap()
        .call_resumable(
            &mut store,
            &[wasmi::Value::I32(0), wasmi::Value::I32(0)],
            &mut out,
        )
        .unwrap();

    loop {
        match call {
            wasmi::ResumableCall::Resumable(r) => {
                let err = r.host_error().downcast_ref::<InterruptedTrap>().unwrap();
                if err.name == "ext_logging_max_level_version_1" {
                    call = r
                        .resume(&mut store, &[wasmi::Value::I32(0)], &mut out)
                        .unwrap()
                } else {
                    println!("{:?}", err.name);
                    println!("success!");
                    break;
                }
            }
            wasmi::ResumableCall::Finished => unreachable!(),
        }
    }
}
```

I have attached the wasm data here: [wasm_data.zip](https://github.com/smol-dot/smoldot/files/10716423/wasm_data.zip) (zipped in order to satisfy GitHub)

If I replace the body of the closure with this, then the test succeeds:

```rust
move |_caller, _params, _ret| {
    if name == "ext_logging_max_level_version_1" {
        _ret[0] = wasmi::Value::I32(0);
        Ok(())
    } else {
        Err(wasmi::core::Trap::from(InterruptedTrap {
            name: name.clone(),
        }))
    }
}
```

In other words, if the function directly returns when it is called, then the test works. If, however, the function emits a trap, then is resumed, then the test fails.
This makes me think that there is a bug in the resumable functions feature.

Robbepop:
@tomaka Thanks a lot for digging into the problem and producing a "minimal" test case!

I am going to try to further minimize it for `wasmi` and try to find and fix the issue. :)
Generally, after having implemented so many new features in `wasmi` the focus for the next weeks is to stabilize, fix bugs, create more tests, improve CI and fuzzing etc.

Robbepop:
@tomaka I fixed the bug in this PR: https://github.com/paritytech/wasmi/pull/671
Please test yourself that this bug is fixed and please check if now smoldot works with the fixed `wasmi` version.
The fix was very simple fortunately.

tomaka:
Seems to work, thank you!

Robbepop:
@tomaka I just release `wasmi v0.26.1` that includes the bug fix to unblock you.

See here what happened between `0.25.0` and `0.26.1`:
- `v0.26.0`: https://github.com/paritytech/wasmi/releases/tag/v0.26.0
- `v0.26.1`: https://github.com/paritytech/wasmi/releases/tag/v0.26.1

Robbepop:
@tomaka btw. if you want to run `wasmi` on steroids it is important that you compile it with:
```toml
[profile.release]
lto = "fat"
codegen-units = 1
```

tomaka:
> @tomaka btw. if you want to run `wasmi` on steroids it is important that you compile it with:
> 
> ```toml
> [profile.release]
> lto = "fat"
> codegen-units = 1
> ```
> 
> makes a huge difference for `wasmi`, especially the newer versions of it.

It's already the case: https://github.com/smol-dot/smoldot/blob/3b8f08b94bc30592aa6ad951c1d7e13ffe4ef06a/Cargo.toml#L128-L136

Robbepop:
@tomaka Do you have any benchmark comparison between old `wasmi` and new `wasmi` for smoldot? I'd be very interested in.

Also another question that was brought up by @athei was if it was tested and verified that smoldot is can still properly sync relay chain from genesis with the new `wasmi`.

athei:
> Disabling all Wasm features (such as `bulk-memory`) also doesn't change anything, which makes me think that the runtime doesn't any these features.

You are correct. The runtime (and PvF) are restricted to the wasm MVP for now. AFAIK the only exception is mutable globals. But don't nail me down on this. But most def nothing as big as bulk memory.

> Also another question that was brought up by @athei was if it was tested and verified that smoldot is can still properly sync relay chain from genesis with the new wasmi.

I brought this up because substrate itself will most likely never upgrade to a newer wasmi. We doubled down on wasmtime and will probably simplify the code soon my removing swappable executor support. So smoldot right now would be the only piece of software which we can use to do a `burn-in` using wasmi (sync a big chain from genesis). Maybe having a manually triggerable CI job in this repo would be a nice addition. However, it is probably not possible to run something for days using gha.

tomaka:
> @tomaka Do you have any benchmark comparison between old wasmi and new wasmi for smoldot? I'd be very interested in.

No. I don't benchmark this because the light client has no performance issue at all w.r.t. running the runtime. Even if wasmi was ten times slower it would still be fine.

> Also another question that was brought up by @athei was if it was tested and verified that smoldot is can still properly sync relay chain from genesis with the new wasmi.

Haven't tested either, as the light client doesn't sync from the genesis, and the full node uses wasmtime by default. Syncing from genesis with smoldot takes a long time and might stall due to unfixed networking issues, so I'm not very motivated to try it out. However, I've just tried syncing a bit around block 5 million (where my database is) and everything seems to work.

tomaka:
Also, smoldot can only sync up to around block 6 million or so if I remember correctly. At some point, the runtime starts using features that smoldot doesn't implement yet.

athei:
> Haven't tested either, as the light client doesn't sync from the genesis, and the full node uses wasmtime by default. Syncing from genesis with smoldot takes a long time and might stall due to unfixed networking issues

Sorry for my ignorance: Those network issues arise when you use smoldot as a full node with wasmi? Sorry if this didn't make sense. I didn't look into smoldot.

> At some point, the runtime starts using features that smoldot doesn't implement yet (code substitutes, state_version v1, and child tries come to mind).

Child tries are pretty old. But were only used for contracts which are not on the relay chain. But I think they were used for crowdloans and this is probably when they become a problem for smoldot. But I get it. It adds a lot of API surface when this could have been implemented with prefix trees as well.

tomaka:
> Sorry for my ignorance: Those network issues arise when you use smoldot as a full node with wasmi? Sorry if this didn't make sense. I didn't look into smoldot for some time.

Completely orthogonal to wasmi.

PR #142, opened 2023-02-09
fix random

turuslan:
wasm smoldot node always started with same peer id  
because [`.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice) (clone) and [`.subarray()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray) (mutable reference) were confused  
so randomness was not written back to wasm memory

- https://github.com/paritytech/smoldot/pull/3095#issuecomment-1424384728

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


tomaka:
Thank you!

This is why I hate JavaScript. I remember thinking about this, but [the documentation](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice) says "a shallow copy". What the hell does "a shallow copy" even mean in the context of `Uint8Array` if not that you're sharing the underlying buffer? Anyway.

tomaka:
I've taken the liberty to push a commit with an update to the CHANGELOG.

tomaka (approved):

tomaka (approved):

Issue #166, opened 2023-02-11
Implement support for child tries


tomaka:

tomaka:
The next step in adding support would be to update [`runtime_host.rs`](https://github.com/smol-dot/smoldot/blob/a9c9e5b590b4e2f783ce404221d8604ab8efc2c0/lib/src/executor/runtime_host.rs#L84).

This basically contains the logic of the pending changes that have been made by the storage but not saved yet.
It also contains the logic of "transactions", where the runtime can start a transaction and then later potentially revert it.

Adding support for child tries in an efficient way (I'm not looking for the most optimal implementation, but at least the best possible `O`-notation complexity) seems extremely complicated, although I haven't precisely looked into it yet.

cmichi:
@tomaka Would you be open to writing out a W3F grant for someone to implement this?

For context: We need this not only for using light clients with smart contract Dapps, but also for some more advanced testing tooling and some features that auditors of ink! contracts requested.

tomaka:
Writing a grant would probably take me way more time than doing it myself

cmichi:
I was honestly hoping for this response 😜 .

cmichi:
So do you think there's a chance of implementing it some time in the next weeks?

tomaka:
Honestly the biggest obstacle is that the way child tries work is a bit of an illogical shit show, and I have the biggest trouble in the world finding someone who is able to explain me how they work in Substrate.

tomaka:
https://github.com/smol-dot/smoldot/issues/272 should to be tackled first in order to avoid having to maintain a cache of each child tries. Removing the caches would considerably simplify the implementation.
An alternative could be to just not have a cache, but that would not be a long term solution. Having no cache first would just make #272 more difficult to do.

tomaka:
I've noticed a bunch of people follow this issue, so to keep you updated:
#639 was a major change blocking this. Now that it has landed, it should be possible to add proper support for child tries in `runtime_host.rs`.

tomaka:
If someone is willing to help, having test cases would be extremely appreciated. Since I'm unfamiliar with runtime development and Substrate APIs in general, just creating a contracts chain would probably take me ten times the time necessary to implement the feature. And given that the way child tries work is very poorly documented, these test cases would be extremely useful.

[...]

PR #167, opened 2023-02-11
Add child tries support in host.rs

tomaka:
cc #166 

This PR is the first step towards support for child tries in the VM.

While in Substrate child tries support is I believe handled by the "storage layer", in smoldot I prefer to make things clear and explicit.
For example, if the Wasm code reads the storage key `:child_storage:default:foo`, I don't emit a `ExternalStorageGet` variant but instead a `ExternalStorageRoot` variant with "foo" as the trie. Similarly, if the Wasm code calls `ext_storage_next_key_version_1` with `:child_storage:default:foo`, I emit a `ExternalStorageNextChildTrie` variant.

Only default child storages are supported, as this is the only kind of child tries that exist at the moment (and in the foreseeable future).

After this PR, child tries don't work yet. There is a lot of effort remaining to implement support for them especially in `runtime_host.rs` (which is the code that handles the storage diff while the call is in progress, including things like storage transactions). However, this PR does a big chunk of the work by designing an API for them.


tomaka:
Note that the design of child tries is a complete clusterfuck full of unspecified corner cases.

PR #194, opened 2023-02-20
Move things around

tomaka:
This PR is relatively big but straight-forward.

It moves the `src` directory to `lib/src`, and all the subdirectories of `bin` to the root. The chain specs have been moved to a new directory called `demo-chain-specs`.

I think that this "flat" layout of the various components of the project is better for the comprehension of people who aren't familiar with the source code.


tomaka:
Due to the way the `wasm-node-size-diff` CI stage works (it checks out the base branch and compiles it), this change will be done in two steps. Another PR will be opened afterwards to remove one remaining `/bin`

PR #203, opened 2023-02-21
Add tests to host.rs

tomaka:
This PR adds a bunch of tests to the `host.rs` module.
The goal is to test whether the core functionalities work as expected and that all host functions work properly.


tomaka:
Merging, even though the tests are very incomplete, because tests are better than no tests.

Issue #62, opened 2023-02-09
Error while executing Wasm VM Trap MemoryAccessOutOfBounds 


tomaka:
https://github.com/paritytech/smoldot/issues/2702

tomaka:
I'm going to close #205, but this should now be fixed after https://github.com/smol-dot/smoldot/pull/210 and https://github.com/smol-dot/smoldot/pull/211

PR #205, opened 2023-02-21
Zero the Wasm VM memory after a reset

tomaka:
This simple change hasn't been done before because it slows down the execution of the full node by a lot.
However, correctness is more important than speed.

I've now added a test to make sure that this behavior stays, and we'll optimize the code later.


tomaka:
Close https://github.com/smol-dot/smoldot/issues/62

tomaka:
This breaks major syncing the full node, as the wasmtime implementation is a bit broken.
Let's not merge just yet.

tomaka:
Superceded by #210 and #211

PR #208, opened 2023-02-22
Properly reinitialize the instance after a reset in wasmtime

tomaka:
Make sure to recreate the `Instance` when `into_prototype` is called.
This was the case for wasmi but not for wasmtime.

cc https://github.com/smol-dot/smoldot/issues/17


tomaka:
This PR unfortunately creates a panic after 10000 blocks because of https://github.com/smol-dot/smoldot/issues/17#issuecomment-1439713819

PR #209, opened 2023-02-22
Switch to node-version: 14 on CI

tomaka:
Version 12 seems more or less deprecated


tomaka:
Should unlock https://github.com/smol-dot/smoldot/pull/201

PR #246, opened 2023-02-28
catch connect error

turuslan:
Deno does not allow unhandled promise rejection and terminates with `error: Uncaught (in promise)`.

Added `.catch` to `Deno.connect`.

It calls `config.onConnectionReset`, like `WebSocket` (success is `.onopen` (then may `.onerror`) then `.onclose`, failure is `.onerror` then `.onclose`).


wasm-node/javascript/src/index-deno.ts
```diff
@@ -224,7 +228,7 @@ function connect(config: ConnectionConfig, forbidTcp: boolean, forbidWs: boolean
             } else {
                 // TCP
                 connection.socket.destroyed = true;
-                connection.socket.inner.then((connec) => connec.close());
+                connection.socket.inner.then((connec) => connec?.close());
```
tomaka, line 231: Can `connec` actually be null/undefined here? I feel like the only way that this could happen is if `reset` was called twice, which would be a smoldot bug

tomaka (commented):
Thanks!

wasm-node/javascript/src/index-deno.ts
turuslan: I returned `null` when `Deno.connect` returns error, so added this null-check

tomaka: Ah I see, I had to re-read the documentation of `Promise.catch` to understand

I'm trying very hard to not lose track of what's happening, which can very easily happen with JavaScript.
Given that `reset` should never be called after `onConnectionReset`, this should never be `null`, so:

```suggestion
                connection.socket.inner.then((connec) => connec!.close());
```

wasm-node/javascript/src/index-deno.ts
```diff
@@ -161,6 +161,10 @@ function connect(config: ConnectionConfig, forbidTcp: boolean, forbidWs: boolean
             inner: Deno.connect({
                 hostname: tcpParsed[2],
                 port: parseInt(tcpParsed[3]!, 10),
+            }).catch((error) => {
+                socket.destroyed = true;
+                config.onConnectionReset(error.toString());
+                return null!;
```
tomaka, line 167: Could you maybe move this as the second parameter of the `then` below?
This would guarantee that the first closure of the `then` never gets a `null`

Also:

```suggestion
            }).catch((error) => {
                socket.destroyed = true;
                config.onConnectionReset(error.toString());
                return null;
```

turuslan: That first closure (and `.send` too) checks `socket.destroyed`, so doesn't use `null`.

Either `null!` is required, or change of type to `TcpConnection { socket(inner: Promise<Deno.TcpConn | null>) }`

tomaka: > or change of type to TcpConnection { socket(inner: Promise<Deno.TcpConn | null>) }

Well, yes, that option.
As far as I understand, `null!` lies to the type checker. It's an escape hatch for situations where TypeScript isn't good enough, and not something you're supposed to use normally.

tomaka:
I've pushed commits doing the change I've requested, plus a CHANGELOG entry

tomaka:
Thanks!

tomaka (approved):

PR #262, opened 2023-03-03
Deduplicate the trie node encoding and decode code

### LINKED ISSUE #45: The trienodevalue and trieproofnodecodec modules are kind of redundant
https://github.com/paritytech/smoldot/issues/3031

tomaka:
Close #45

There are currently two modules that are extremely similar: `node_value` and `proof_node_codec`.
This is because a node in a proof is the same encoding as a node in the trie. A proof is simply a subset of the trie.

This PR merges them into a new `trie_node.rs` module.

`trie_node.rs` is identical to `proof_node_codec.rs` except that:

- `Decoded` now accepts a second template parameter for the type of the slice of the children.
- It has the `calculate_merkle_value` function of `node_value`, and `node_value::Output` is now `trie_node::MerkleValueOutput`.


lib/src/trie/trie_node.rs
```diff
@@ -157,10 +165,138 @@ pub fn encode_to_vec(
...
+pub fn calculate_merkle_value(
+    decoded: Decoded<
+        '_,
+        impl ExactSizeIterator<Item = nibble::Nibble> + Clone,
+        impl AsRef<[u8]> + Clone,
+    >,
+    is_root_node: bool,
+) -> Result<MerkleValueOutput, EncodeError> {
+    // The partial key can only be empty if the node is the root.
+    debug_assert!(is_root_node || decoded.partial_key.len() != 0);
```
tomaka, line 185: This assertion is triggered, which seems to indicate a pretty nasty issue in `trie_structure.rs`

tomaka: Ok nevermind, I'm an idiot, this is in fact completely valid.

PR #276, opened 2023-03-10
make runtime maximum log level configurable

ermalkaleci:


tomaka:
Thanks for the PR

- I'd actually prefer to keep this as an opaque value. The fact that for example "4" means "debug" is specific to the runtime. There's nothing that says that "debug" can't be "90000" for example. I think we can mention that for example 4 generally means debug, but make it clear that it's not necessarily true.

- The same change should be done for `read_only_runtime_host` as well.

- The configuration option should be propagated, for example in `author.rs` you should also give the option to make the log level configurable. Or at the very least a TODO should be added.

ermalkaleci:
Yeah true, this value will depend on runtime implementation and "4" may not be "debug"

lib/src/executor/read_only_runtime_host.rs
```diff
@@ -339,6 +344,8 @@ struct Inner {
     vm: host::HostVm,
     /// Concatenation of all the log messages generated by the runtime.
     logs: String,
+    /// Runtime maximum log level
+    pub max_log_level: u32,
```
tomaka, line 348:
```suggestion
    max_log_level: u32,
```

lib/src/executor/read_only_runtime_host.rs
```diff
@@ -339,6 +344,8 @@ struct Inner {
     vm: host::HostVm,
     /// Concatenation of all the log messages generated by the runtime.
     logs: String,
+    /// Runtime maximum log level
```
tomaka, line 347:
```suggestion
    /// Value provided by [`Config::max_log_level`].
```

lib/src/author/runtime.rs
```diff
@@ -449,6 +454,8 @@ struct Shared {
     block_body: Vec<Vec<u8>>,
     /// Concatenation of all logs produced by the multiple calls.
     logs: String,
+    /// Runtime maximum log level
```
tomaka, line 457:
```suggestion
    /// Value provided by [`Config::max_log_level`].
```

lib/src/author/build.rs
```diff
@@ -229,6 +229,7 @@ impl AuthoringStart {
                     })
                 }
             },
+            max_log_level: Default::default(),
```
tomaka, line 232: This should be a field in the `Config` struct found in `build.rs`

lib/src/chain/chain_information/build.rs
```diff
@@ -391,6 +391,7 @@ impl ChainInformationBuild {
                 function_to_call: call.function_name(),
                 parameter: call.parameter_vectored(),
                 virtual_machine: inner.virtual_machine.take().unwrap(),
+                max_log_level: Default::default(),
```
tomaka, line 394:
```suggestion
                max_log_level: 0,
```

light-base/src/sync_service/parachain.rs
```diff
@@ -1127,6 +1127,7 @@ async fn parahead<TPlat: Platform>(
             parachain_id,
             para::OccupiedCoreAssumption::TimedOut,
         ),
+        max_log_level: Default::default(),
```
tomaka, line 1130:
```suggestion
        max_log_level: 0,
```

lib/src/author/runtime.rs
```diff
@@ -91,6 +91,10 @@ pub struct Config<'a> {
     /// Capacity to reserve for the number of extrinsics. Should be higher than the approximate
     /// number of extrinsics that are going to be applied.
     pub block_body_capacity: usize,
+
+    /// Runtime maximum log level
+    /// 0 means off, 1 means error, 2 means warn, 3 means info, 4 means debug, 5 means trace.
```
tomaka, line 96:
```suggestion
    /// Maximum log level of the runtime.
    ///
    /// > **Note**: This value is opaque from the point of the view of the client, and the runtime is free to interpret it the way it ways. However, usually values are: `0` for "off", `1` for "error", `2` for "warn", `3` for "info", `4` for "debug", and `5` for "trace".
```

lib/src/executor/read_only_runtime_host.rs
```diff
@@ -35,6 +35,10 @@ pub struct Config<'a, TParams> {
     /// Parameter of the call, as an iterator of bytes. The concatenation of bytes forms the
     /// actual input.
     pub parameter: TParams,
+
+    /// Runtime maximum log level
+    /// 0 means off, 1 means error, 2 means warn, 3 means info, 4 means debug, 5 means trace.
```
tomaka, line 40:
```suggestion
    /// Maximum log level of the runtime.
    ///
    /// > **Note**: This value is opaque from the point of the view of the client, and the runtime is free to interpret it the way it ways. However, usually values are: `0` for "off", `1` for "error", `2` for "warn", `3` for "info", `4` for "debug", and `5` for "trace".
```

lib/src/executor/runtime_host.rs
```diff
@@ -72,6 +72,10 @@ pub struct Config<'a, TParams> {
     /// Initial state of [`Success::offchain_storage_changes`]. The changes made during this
     /// execution will be pushed over the value in this field.
     pub offchain_storage_changes: storage_diff::StorageDiff,
+
+    /// Runtime maximum log level
+    /// 0 means off, 1 means error, 2 means warn, 3 means info, 4 means debug, 5 means trace.
```
tomaka, line 77:
```suggestion
    /// Maximum log level of the runtime.
    ///
    /// > **Note**: This value is opaque from the point of the view of the client, and the runtime is free to interpret it the way it ways. However, usually values are: `0` for "off", `1` for "error", `2` for "warn", `3` for "info", `4` for "debug", and `5` for "trace".
```

lib/src/executor/runtime_host.rs
```diff
@@ -631,6 +636,9 @@ struct Inner {
 
     /// Concatenation of all the log messages generated by the runtime.
     logs: String,
+
+    /// Runtime maximum log level
```
tomaka, line 640:
```suggestion
    /// Value provided by [`Config::max_log_level`].
```

lib/src/executor/runtime_host.rs
```diff
@@ -944,7 +951,7 @@ impl Inner {
                             Ok(())
                         }
                     }
-                    match fmt::write(&mut WriterWithMax(&mut self.logs), format_args!("{req}")) {
+                    match fmt::write(&mut WriterWithMax(&mut self.logs), format_args!("{req}\n")) {
```
tomaka, line 954: That doesn't seem right?

```suggestion
                    match fmt::write(&mut WriterWithMax(&mut self.logs), format_args!("{req}")) {
```

A `LogEmit` is emitted when the runtime prints out *something*, not necessarily a line. The runtime has also host functions available that print just a number for example.

tomaka (changes requested):

lib/src/executor/runtime_host.rs
ermalkaleci: Yeah but this will merge logs together into one line and sometimes one word without any space

lib/src/author/build.rs
ermalkaleci: thought we don't need it in this case, adding it

full-node/src/run/consensus_service.rs
```diff
@@ -693,6 +693,7 @@ impl SyncBackground {
                     parent_runtime,
                     block_body_capacity: 0, // TODO: could be set to the size of the tx pool
                     top_trie_root_calculation_cache: None, // TODO: pretty important for performances
+                    max_log_level: 0, // TODO: propagate config, not sure which is the best way
```
tomaka, line 696:
```suggestion
                    max_log_level: 0,
```

This is fine as is

lib/src/author/runtime.rs
```diff
@@ -91,6 +91,11 @@ pub struct Config<'a> {
     /// Capacity to reserve for the number of extrinsics. Should be higher than the approximate
     /// number of extrinsics that are going to be applied.
     pub block_body_capacity: usize,
+
+    /// Maximum log level of the runtime.
+    ///
+    /// > **Note**: This value is opaque from the point of the view of the client, and the runtime is free to interpret it the way it ways. However, usually values are: `0` for "off", `1` for "error", `2` for "warn", `3` for "info", `4` for "debug", and `5` for "trace".
```
tomaka, line 97:
```suggestion
    /// > **Note**: This value is opaque from the point of the view of the client, and the runtime is free to interpret it the way it wants. However, usually values are: `0` for "off", `1` for "error", `2` for "warn", `3` for "info", `4` for "debug", and `5` for "trace".
```

Sorry

lib/src/executor/runtime_host.rs
tomaka: That still doesn't make it right. Plus, it's off topic for this PR and should be in a different PR.

tomaka: https://github.com/smol-dot/smoldot/pull/283

lib/src/verify/header_body.rs
```diff
@@ -364,6 +364,7 @@ pub fn verify(
             top_trie_root_calculation_cache: config.top_trie_root_calculation_cache,
             storage_top_trie_changes: Default::default(),
             offchain_storage_changes: Default::default(),
+            max_log_level: 0,
```
tomaka, line 367: Well, this also needs to be a config option

lib/src/transactions/validate.rs
```diff
@@ -327,6 +327,7 @@ pub fn validate_transaction(
                 top_trie_root_calculation_cache: None,
                 storage_top_trie_changes: storage_diff::StorageDiff::empty(),
                 offchain_storage_changes: storage_diff::StorageDiff::empty(),
+                max_log_level: 0,
```
tomaka, line 330: And this

tomaka (approved):
Thanks!

ermalkaleci:
Thank you! Chopsticks is possible thanks to smoldot :)

PR #284, opened 2023-03-12
Merge the two subscription hashmaps into one

tomaka:
This is the first step towards the simplification of the clusterfuck that is the current JSON-RPC server code.

This PR merges the two hashmaps `misc` and `chain_head_follow` into a single one.

The next step will be to move this hashmap into the `requests_subscriptions` state machine, which should simplify things a lot.

As part of this PR, I've also fixed a bug that caused the chainHead_follow subscriptions from not being properly cleaned up. I've also removed a lot of calls to `str::to_owned()` by using `HashMap::entry_ref`.

The diff in `chain_head.rs` is unfortunately pretty large because `rustfmt` has decided to reformat entire blocks, but apart from that bugfix all the changes are insubstantial.

### DIFF
--- a/light-base/src/json_rpc_service/chain_head.rs
+++ b/light-base/src/json_rpc_service/chain_head.rs
@@ -29,7 +29,7 @@ use core::{
     time::Duration,
 };
 use futures::prelude::*;
-use hashbrown::HashMap;
+use hashbrown::{hash_map, HashMap};
 use smoldot::{
     chain::fork_tree,
     executor::{self, runtime_host},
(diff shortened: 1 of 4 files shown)


light-base/src/json_rpc_service/chain_head.rs
```diff
@@ -1379,14 +1405,24 @@ impl<TPlat: Platform> Background<TPlat> {
...
+                Some(state_machine_subscription)
+            }
+            hash_map::EntryRef::Occupied(_) | hash_map::EntryRef::Vacant(_) => None,
+        };
+
+        if let Some(state_machine_subscription) = &state_machine_subscription {
+            self.requests_subscriptions
+                .stop_subscription(state_machine_subscription)
+                .await;
         }
```
tomaka, line 1426: This is the bugfix I mention in the OP. This block didn't exist before.

Issue #291, opened 2023-03-14
Refactoring `light-base/src/json_rpc_service.rs`


tomaka:
The code in `json_rpc_service.rs` is way too cumbersome and complicated, and more functionalities should be added to the `request_subscriptions` state machine in order to make it more simple.

Unfortunately, the logic of this code *is* inherently complicated. When a `chainHead_unstable_body` for example is called, we need to look up in the corresponding follow subscription to check whether the block is still pinned.

I can see two ways of making it work:

- `request_subscriptions` could store, for each subscription, a task and a user data. When the user wants to unsubscribe, we ask the `request_subscriptions` to abort the task (after examining the user data, to be sure that the type of the subscription matches). The user data is behind a `Mutex` and would store, in the case of a follow subscription, the `FollowSubscription` struct that currently exists. This solution is pretty similar to what already is done in the code.

- `request_subscriptions` could store, for each subscription, a task and a channel that can send messages to that task. The channel would be provided by the `request_subscriptions` rather than done manually in order to add some safety around locking `Mutex`es. When the user wants to unsubscribe, we send a message to the subscription in question which then politely shuts down. When the user calls for example `chainHead_unstable_body`, we send a message to the subscription in question as well, which then starts the body request subscriptions. The existing `FollowSubscription` would be a local variable in the task of the follow subscription.

The latter is more safe in terms of `Mutex` locking, as it would prevent all accidental deadlocks, but maybe also a bit more complicated.

tomaka:
cc https://github.com/smol-dot/smoldot/pull/284 which was the first step

tomaka:
I'm leaning towards the second solution, as the lack of mutexes seems more clean to me.

tomaka:
Done after https://github.com/smol-dot/smoldot/pull/793

PR #298, opened 2023-03-15
Abort subscriptions using messages rather than abort handles

tomaka:
cc https://github.com/smol-dot/smoldot/issues/291

This relatively big PR reworks some aspects of how JSON-RPC subscriptions work:

- We no longer abort tasks dedicated to subscriptions when asked by the user to cancel them.
- Subscriptions now have an MPSC channel (with a buffer of only one item) that can be used to send messages to them. These messages are processed by the subscription task.
- When asked to cancel a subscription, we now send a message to the corresponding task in order to ask it to gracefully stop. The subscription sends back a confirmation so that the unsubscription request handler knows if the subscription still exists and is of the correct type.
- Similarly, the chainHead_body/call/storage/header/unpin functions now send a message to the task of the corresponding chainHead_follow in order to ask it to perform the action. This allows moving the `FollowSubscription` struct locally in the chainHead_follow task.

While this PR doesn't make the code more simple, it is a preliminary step to moving some functionalities to `request_subscriptions.rs`.

### DIFF
--- a/light-base/src/json_rpc_service.rs
+++ b/light-base/src/json_rpc_service.rs
@@ -61,7 +61,11 @@ use core::{
     sync::atomic,
     time::Duration,
 };
-use futures::{channel::mpsc, lock::Mutex, prelude::*};
+use futures::{
+    channel::{mpsc, oneshot},
+    lock::Mutex,
+    prelude::*,
+};
 use hashbrown::HashMap;
 use smoldot::{
     chain::fork_tree,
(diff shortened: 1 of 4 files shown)


tomaka:
Everything seems to work on the first try.
The chances that there are bugs somewhere are quite high, however this is true for the existing code as well. The objective of the simplification is precisely to make the code easy to review.

PR #300, opened 2023-03-15
Add a tasks queue to requests_subscriptions

tomaka:
Next step in https://github.com/smol-dot/smoldot/issues/291

This PR adds to the `requests_subscriptions` state machine a queue of tasks. When a subscription is started, its corresponding task is added to the queue. Then, in parallel, some tasks run the subscriptions tasks.

This have the advantage of guaranteeing that the subscriptions tasks do not have higher priority than processing new requests.

The next step will be to merge `start_subscription` with `add_subscription_task` and other things.

Surprisingly, I couldn't find a good library in the ecosystem that does what the `executor` module does, so I wrote an executor myself. Even though it uses unsafe code, I'm very confident in the code of that executor, and the test basically covers every situation.

### DIFF
--- a/lib/src/json_rpc/requests_subscriptions.rs
+++ b/lib/src/json_rpc/requests_subscriptions.rs
@@ -37,7 +37,8 @@
 //!
 //! - One lightweight task for each client currently connected to the server.
 //! - A fixed number of lightweight tasks (e.g. 16) dedicated to answering requests.
-//! - A fixed number of lightweight tasks (e.g. 8) dedicated to processing subscription tasks.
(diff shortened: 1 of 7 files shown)


tomaka:
One issue with this code is that if the task isn't woken up, it leaks even if the `TasksQueue` is destroyed.
I've fixed this by making it less efficient but also by removing the unsafe code.

tomaka:
I've added some TODOs related to future cancellation issues in `executor.rs`.
I have ideas on how to fix them, but they're niche problems and hence fine to leave as TODOs. I'd prefer to advance in the JSON-RPC service simplification at the moment.

tomaka:
Fixed the cancellation issues.

PR #303, opened 2023-03-16
More powerful subscriptions in `requests_subscriptions`

tomaka:
Continuation of https://github.com/smol-dot/smoldot/issues/291

This PR attempts to simplify how subscriptions are implemented in the high-level code by moving some logic to `requests_subscriptions`:

- Each subscription within `RequestsSubscriptions` now has a task associated to it.
- Sending a message to a subscription task is now done with a function on `RequestsSubscriptions`.
- Subscriptions are automatically cleaned up when a subscription task is destroyed, instead of manually.


tomaka:
I'm going to stop here for this PR, despite other simplifications possible. I don't want to clutter one PR too much.

PR #315, opened 2023-03-19
Improve the networking abstraction of the Platform trait

tomaka:
This PR modifies the `Platform` trait exposed by `light-base` in order to improve the abstraction over "streams" (i.e. TCP or WebSocket connections, or WebRTC substreams).

It now allows closing the sending side, and no longer allows an unlimited amount of outgoing data.

As part of this PR, most of the `AsyncStdTcpWebSocket` has been rewritten in a clean way. There's no cursed background task that sends and receives messages from the front or whatnot. Instead, things are now "flat" and the socket is polled directly in the future manipulated by the networking task.

Finishing this consists in updating the wasm-node implementation of `Platform` as well.


tomaka:
While updating the `Platform` implementation in the wasm node, I've discovered several bugs where we call `*_reset` on a connection or stream that has already been reset by the remote. This is now fixed.

Issue #317, opened 2023-03-20
Smoldot has panicked while executing task `connections`


wirednkod:
When trying to test  [this Substrate Connect issue with Astar](https://github.com/paritytech/substrate-connect/issues/1380), with Substrate Connect Extension (including Smoldot v.1.0) the following errors appear:

```
Smoldot has panicked while executing task `connections`. This is a bug in smoldot. Please open an issue at https://github.com/smol-dot/smoldot/issues with the following message:
panicked at 'assertion failed: request.len() <= max_size', /__w/smoldot/smoldot/lib/src/libp2p/connection/established/single_stream.rs:801:17
```
and
```
content.js:2 Uncaught Error: panicked at 'assertion failed: request.len() <= max_size', /__w/smoldot/smoldot/lib/src/libp2p/connection/established/single_stream.rs:801:17
    at Object.onWasmPanic (content.js:2:2943130)
    at Object.onPanic (content.js:2:2893904)
    at panic (content.js:2:2894242)
    at 0105409e:0xb137c
    at 0105409e:0x2a470d
    at 0105409e:0x2a3b5b
    at 0105409e:0x2a3ad2
    at 0105409e:0x2a41fb
    at 0105409e:0x2a9f28
    at 0105409e:0x2aa450
```
    
 Replication steps:
1- Make sure latest Substrate Connect Extension is added in browser (v0.2.3);
2-  Open https://staging.portal.astar.network/astar/dashboard and open up the console;
3-  Connect to a wallet (top right) and switch to light client;

![image](https://user-images.githubusercontent.com/5408605/226297081-e8da7f09-b016-4ca8-91f8-19e77dc03e0b.png)

tomaka:
For what it's worth, this is an issue that wouldn't happen if we were using the new JSON-RPC API.

PR #318, opened 2023-03-20
Return an error if a request exceeds the maximum allowed size, instead of panicking

tomaka:
Fix https://github.com/smol-dot/smoldot/issues/317


tomaka:
I'm going to release v1.0.1 with that PR.
I want to release a version before merging https://github.com/smol-dot/smoldot/pull/303, as that is a big "risky" PR.

PR #340, opened 2023-03-23
Check whether document is defined

tomaka:


wasm-node/javascript/src/index-browser.ts
```diff
@@ -51,6 +51,8 @@ export function start(options?: ClientOptions): Client {
       return Promise.resolve(inflate(classicDecode(input)))
     },
     registerShouldPeriodicallyYield: (callback) => {
+      if (!document)  // We might be in a web worker.
```
josepot, line 54: This would still throw inside a WebWorker. The correct way to do it would be:

```suggestion
      if (typeof document === 'undefined')   // We might be in a web worker.
```

josepot (changes requested):

PR #345, opened 2023-03-24
Warn if a task takes a long time to poll

tomaka:
This should help detect small freezes.

Before merging this PR, the freezes should actually be fixed, so as to not start printing warnings to the users.


tomaka:
One freeze probably comes from here: https://github.com/smol-dot/smoldot/blob/b7bf5b95a3dcc86476f600a16bf98875fd18ee41/lib/src/sync/all.rs#L1172-L1220

tomaka:
When in the browser, this prints a lot of warnings about tasks taking exactly 16ms to poll. I have no idea why.
It could be the case that `performance.now()` has a low precision, but the docs mention 5µs or 100µs, which is pretty far from 16ms: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#security_requirements

tomaka:
`state_getKeysPaged` seems to take a long time to decode the proof and run the `prefix_proof` system.
It seems that decoding a proof should be split into stages.

tomaka:
Verifying a finality proof should also be split into stages.

tomaka:
I'm going to merge this PR, but not mention that an issue should be opened.

Issue #44, opened 2023-02-09
The network key should periodically change


tomaka:
https://github.com/paritytech/smoldot/issues/3077

tomaka:
Right now, it is intended that the network key gets regenerated every time the light client is loaded. This doesn't work due to the issue found by https://github.com/paritytech/smoldot/pull/3095

However, this is IMO not enough. Smoldot could in principle run for days or weeks, in which case the network key should indeed get periodically regenerated.

tomaka:
The modifications needed for this issue probably overlap with https://github.com/smol-dot/smoldot/issues/111

tomaka:
I can see three ways of implementing the rotation:

- Rotate the network key if we hit 0 peers, meaning that it would rotate for example if you lose Internet connectivity for a whole or if you put your laptop in sleep mode.
- Rotate the network key at a periodic interval, and when doing so kill all existing connections.
- Allow a mix of connections with multiple different local PeerIds, and periodically rotate the network key used for newly-established connections only, without touching the existing connections.

As a reminder, the objective is to avoid giving the possibility for full nodes to maintain a map of `IP address <=> PeerId`, since that would allow tracking users if they move around the planet.

Solution 3 seems strictly superior in all ways to solution 1, assuming that the key rotation period is small enough to not give the time for the user to travel somewhere and maintain the same key. We could also do solution 3, but rotate the key both periodically and when we hit 0 peers.

One thing that could unfortunately happen is: the user is on a fixed location IP that is geotracable, then switches to mobile IP while keeping the same PeerId, then switches again to another fixed location IP that is geotracable while keeping the same PeerId. Even if the PeerId rotates, as long as it doesn't rotate precisely when the IP address changes, the full node can trace the movement between the two fixed location IPs.

However I think that we can simply ignore this corner case, and do this on a best effort rather than reliably.

tomaka:
One tricky issue I didn't think of is how to handle the Kademlia protocol. This is to be brain stormed.

tomaka:
> One tricky issue I didn't think of is how to handle the Kademlia protocol. This is to be brain stormed.

I think that it's actually a non-issue.
Light clients rotate their key but don't need to have k-buckets, and full nodes don't rotate their key.

tomaka:
After https://github.com/smol-dot/smoldot/pull/1255 this is done for the light client.
[...]

PR #377, opened 2023-03-30
Store the `NoiseKey` in `peers.rs` instead of `collection.rs`

tomaka:
Step towards https://github.com/smol-dot/smoldot/issues/44

Right now, the `NoiseKey` is stored in `collection.rs` within an `Arc`. This `Arc` is shared with all the background tasks.

This PR removes the `NoiseKey` from `collection.rs`, and you must now pass a `&NoiseKey` every time you add a connection to the collection. This allows using multiple different noise keys if desired.

The `NoiseKey` is now stored in `peers.rs`. Eventually, the plan is to move it to `service.rs` or even outside.
Unfortunately, `peers.rs` might be a bit of a blocker, because the same remote peer might open multiple substreams towards the multiple local identities without it being a protocol violation.
Even if `peers.rs` turns out to be a blocker, the change in this PR is good even in isolation.

As part of this change, we now provide the `NoiseKey` immediately when initializing a single stream handshake, so that we can pass a `&NoiseKey` by reference at initialization rather than having to store it by value and load it later.

### DIFF
--- a/fuzz/fuzz_targets/network-connection-raw.rs
+++ b/fuzz/fuzz_targets/network-connection-raw.rs
@@ -40,7 +40,6 @@ libfuzzer_sys::fuzz_target!(|data: &[u8]| {
             // This timeout doesn't matter as we pass dummy time values.
             handshake_timeout: Duration::from_secs(5),
             ping_protocol: "ping".into(),
-            noise_key: smoldot::libp2p::connection::NoiseKey::new(&[0; 32]),
         });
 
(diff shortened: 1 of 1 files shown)


tomaka:
I'm pushed a small side-change that replaces `Box<NoiseHandshake>` with `NoiseHandshake`.
This is because `NoiseHandshake` now contains a `Box<NoiseHandshakeInner>` rather than the fields directly.

Issue #60, opened 2023-02-09
Brainstorm and refactor the discovery process and address book


tomaka:
https://github.com/paritytech/smoldot/issues/2746

tomaka:
I'm going to close this as done after https://github.com/smol-dot/smoldot/pull/1200 and https://github.com/smol-dot/smoldot/pull/1269.

PR #391, opened 2023-04-04
Properly implement connection shutdowns in collection.rs

tomaka:
Very work in progress


tomaka:
While the changes in `collection.rs` are (tough but) mechanical, I'm unsure how to propagate these changes to `peers.rs` and `service.rs`. At the moment, they aren't capable of handling half-closed connections. In particular, `service.rs` automatically unallocates the peer's slot when its connection is shutting down.

Maybe doing https://github.com/smol-dot/smoldot/issues/60 first is a better idea.

tomaka:
This can now be worked upon again after #1200. However I'll most likely open a new PR.

Issue #350, opened 2023-03-27
Enable some Wasm features when compiling the wasm node


tomaka:
`bulk-memory` comes to mind, but there might be others.

Of course, the blocker is that all major browsers should support a feature before it can be enabled.

tomaka:
https://github.com/smol-dot/smoldot/pull/356 enables `bulk-memory` and `signext`.

The two others that would be appropriate are:

- `simd128`. Unfortunately, this is only supported by NodeJS 16.4, which I feel is a bit too recent.
- `multivalue`. Enabling this one crashes rustc.

tomaka:
> multivalue. Enabling this one crashes rustc.

It seems that `rustc` no longer crashes, but I'm getting linking errors between the `std` library and my code, due to the fact that the standard library that is shipped with `rustup` isn't compiled with this option.

tomaka:
Overall, `multivalue` doesn't seem important in our case, so I'm going to close this issue as completed.

PR #395, opened 2023-04-06
Bump minimum version to NodeJS v16 and enable SIMD

tomaka:
cc https://github.com/smol-dot/smoldot/issues/350

We now require NodeJS v16.

Since NodeJS v14 is deprecated only at the end of the month, what this PR does is now 100% accurate but will be accurate at the end of the month.


tomaka:
It seems that `twiggy` doesn't like SIMD operations.

See https://github.com/rustwasm/twiggy/issues/632, which would be fixed by https://github.com/rustwasm/twiggy/pull/654. This small PR has been opened for 9 months and not merged yet, which probably shows that it's not going to be merged anytime soon.

Issue #272, opened 2023-03-07
RPC call returns "PrefixKeysForbidden" error code


skunert:
Runtime calls of the `persisted_validation_data` method via `state_call` return a "PrefixKeysForbidden" error. 

Previously I assumed this is related to https://github.com/smol-dot/smoldot/issues/177, since this runtime call uses the `state_root` host function. However, should the execution proof that the full node delivers not include all required state entries? Why do we need this prefix key fetching here? Is it a bug after all?

tomaka:
An unintended side effect of https://github.com/smol-dot/smoldot/pull/259 is that the runtime code might request the list of all keys in the entire storage in order to build the trie root cache. This might be why you're getting this error, and this is definitely problematic.

However, as a side note, I really feel like there's a problem somewhere with what you're doing. `persisted_validation_data` really isn't supposed to do anything else but read the storage.
Smoldot itself calls this function when you connect to a parachain, and everything works fine and has worked long before #259.

skunert:
The problem can be reproduced by using this json rpc request a polkadot light client:
```json
{"id": "id", "method": "state_call", "params": ["ParachainHost_persisted_validation_data", "0xe803000001"], "jsonrpc": "2.0"}
```
This is be a valid json rpc call for polkadot and should request the persisted validation data for statemint.
`e803000001` = `(ParaId(1000), TimedOut)`.


Looking at the runtime call implementation [here](https://github.com/paritytech/polkadot/blob/d0419a924137f4b15ec9a2e76d2a44cc6330ce42/runtime/parachains/src/runtime_api_impl/v2.rs#L213-L225) and [here](https://github.com/paritytech/polkadot/blob/d0419a924137f4b15ec9a2e76d2a44cc6330ce42/runtime/parachains/src/runtime_api_impl/v2.rs#L187-L210), the call is indeed read-only when the occupied core assumption is `TimedOut`.


Also you are right that this is related to #259. Before that commit, the call works fine. After, it is broken. This is also why the smoldot internal calls succeed, they still use the read-only-runtime-host if I see that correctly.

tomaka:
Ah ok, I had wrongly assumed that `ParachainHost_persisted_validation_data` was also failing when you reported https://github.com/smol-dot/smoldot/issues/254

[...]

PR #425, opened 2023-04-14
Add `TrieStructure::range`

tomaka:
cc https://github.com/smol-dot/smoldot/issues/272

While working on https://github.com/smol-dot/smoldot/pull/424, I realized that I needed a `range` function for `TrieStructure`.

However, it turns out that this is insanely complicated to implement properly.

I've started this PR but I'm going to abandon it and implement `PrefixKeys` inefficiently, which is ok since `PrefixKeys` is going to be removed in the future.

I'm opening this PR in order to not lose the code in case this function ends up being needed, but I'm not planning to finish it at the moment.

### DIFF
--- a/lib/src/trie/nibble.rs
+++ b/lib/src/trie/nibble.rs
@@ -27,10 +27,13 @@ impl Nibble {
         Nibble(0)
     }
 
+    /// Returns the equivalent of `Nibble::try_from(15).unwrap()`. It is the maximum possible value
+    /// for a nibble.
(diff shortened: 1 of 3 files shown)


tomaka:
It turns out that this function really is the easiest way to do things.
I've now finished implement it, but it probably unfortunately has loads of bugs.
Tests are yet to be written.

tomaka:
I believe that the body of the iterator is now correct, and it's the "start search" phase that is buggy.

Issue #133, opened 2023-02-09
Make the library compile for `#![no_std]`


tomaka:
https://github.com/paritytech/smoldot/issues/106

tomaka:
It should be possible to replace `snow` with the crates of https://github.com/RustCrypto

tomaka:
After #870, I think the only blockers remaining are the crates that are marked with `TODO`s in the `Cargo.toml`, name `ruzstd` (which is waiting for https://github.com/rust-lang/rust/issues/103765), the `futures-*` crates, and `event-listener`.

tomaka:
The last blocker seems to be that `f64::powf` isn't available in no_std contexts.
It should be possible to bypass the issue, but given that the calculation is copy-pasted from Substrate this seems a bit risky.

tomaka:
`smoldot` (the library) will be no-std-friendly after `ruzstd` publishes a version (see https://github.com/smol-dot/smoldot/pull/431)
As for `smoldot-light`, I believe that it just needs the removal of `futures-util`.

PR #431, opened 2023-04-17
Check on CI whether library compiles for no_std

tomaka:
cc https://github.com/smol-dot/smoldot/issues/133

This PR is a revival of https://github.com/paritytech/smoldot/pull/1791

It is not expected to pass CI right now, but it is a long term objective that it does.


tomaka:
It seems that `--no-default-features` gets ignored when compiling from the root of the workspace. I'm pretty sure that's a cargo bug.

tomaka:
Additionally, the features of the `dev-dependencies` are unfortunately leaking in the `--lib` build.

tomaka:
> Additionally, the features of the dev-dependencies are unfortunately leaking in the --lib build.

This can be solved by building `smoldot-light` instead of `smoldot`, except for the fact that `env_logger` isn't no-std and that dev-dependencies can't be `optional` :face_exhaling:

tomaka:
> Additionally, the features of the dev-dependencies are unfortunately leaking in the --lib build.

Should be solved by https://github.com/smol-dot/smoldot/pull/656

tomaka:
According to CI, the poly crate fails to compile for an unknown reason.
If it turns out to be a target-specific bug, I've seen people use `thumbv7m-none-eabi` instead.

tomaka:
I've just pushed a commit that overwrites the version of `ruzstd` from crates.io to use the GitHub version instead. I've done this because I would like to see if there is any other blocker. This should obviously be reverted before merging.

Issue #111, opened 2023-02-09
Use only one network service for all chains


tomaka:
https://github.com/paritytech/smoldot/issues/997

tomaka:
While this would be a nice to have feature, it is unfortunately insanely complicated.

tomaka:
The plan, to implement that, is to no longer pass the list of protocols to `peers.rs` or `collection.rs` (or anything below) ahead of time.

Instead, the parameters of the protocol are now passed every single time a substream is opened. When the remote opens a substream, it generates an event which the `service.rs` must answer, to determine the type and parameters of the protocol requested by the remote.
For this reason, it is easier is https://github.com/smol-dot/smoldot/issues/59 done beforehand.

This makes it possible for `service.rs` to support additional protocols and remove support for existing protocols.

Note that when it comes to `peers.rs`, since the state machine must track duplicate substreams, it is unclear how exactly this would look like at the moment.

tomaka:
Note that it is possible for multiple chains with the same genesis hash and fork ID to be added to the light client.
When that happens, all the chains with the same genesis hash and fork ID should use the same `ChainId` of that networking service.

This has the consequence that the network service must change to be able to report the existing state of the networking, which is something we already want to do anyway for #519.

PR #462, opened 2023-04-23
No longer pass the list of supported protocols to individual connections

tomaka:
This PR is the continuation of https://github.com/smol-dot/smoldot/pull/382, and is more progress towards https://github.com/smol-dot/smoldot/issues/111

Rather than keep the list of known protocols in the connection-dedicated tasks, these tasks now send an event to the coordinator whenever an inbound substream arrives. The coordinator then accepts or denies this substream.
This makes it possible to adjust the list of protocols on the fly. The plan, however, is not to adjust the list of protocols of `collections.rs`, but to move this list of protocols all the way up to `service.rs`.

The main difficulty was how to keep track of which protocol has been accepted. For example, if you receive an inbound substream with `my-awesome-protocol`, and later you receive a request on that substream, you need to remember that it was a request on the `my-awesome-protocol`.
In this PR, this is done through a user data associated with each substream. In order to implement this, I've also merged the `TNotifUd` and `TRqUd` into one and expanded it to inbound substreams.
In the future, however, I would like to track this through a hashmap in the coordinator, and then later in `service.rs`.

Can be reviewed commit by commit.


tomaka:
Overall, I'm really not happy with the way the abstract "substream state machine" works. It is rather complicated to keep track in your head of all the possible states a substream can be in (not negotiated, negotiated, request, notifications out, etc.). I feel like this should be refactored somehow, but I don't really know how to keep this simple.

PR #479, opened 2023-04-26
Substream::reset() is now just a state transition

tomaka:
This PR attempts to simplify `established/single_stream.rs` by removing the special handling of dead substreams.
When a substream is reset by the remote, we now immediately reset it, and its events are processed at the next iteration.


tomaka:
Closing in favor of https://github.com/smol-dot/smoldot/pull/958

Both this PR and https://github.com/smol-dot/smoldot/pull/958 have the objective of simplifying the code, but https://github.com/smol-dot/smoldot/pull/958 does it way better.

PR #481, opened 2023-04-26
Use the wasi functions for time retrieval, and yield less often to the JS engine

### LINKED ISSUE #418: Update the bindings now that BigInt <-> i64 is stable
According to https://webassembly.org/roadmap/, you can now have JS functions matching Wasm functions that have a `i64`.
As such, we can use WASI to retrieve the current time rather than a custom function.

This stability has only been the case since NodeJS 15, so https://github.com/smol-dot/smoldot/pull/395 is necessary.

tomaka:
Fix #418


tomaka:
Everything times out for some reason, and I don't understand why.

tomaka:
I think that it's because it's become so slow that the CPU is completely clogged.

tomaka:
The issue happens in `cpu_rate_limiter.rs`. Switching back to the old `Instant` in only that module fixes the issue.

If I print the value of `poll_duration` before and after, before the value is generally around 100µs and sometimes as low as 6µs. If I print the value after, it is always between 3µs and 7µs.

I'm still completely missing what could be happening.

tomaka:
My hypothesis is that the change in precision in the time causes the CPU limiter to call `setImmediate` very very often, thus preventing any other JS task (such as the connection succeeding) from being executed.

tomaka:
Confirmed that this was the issue.
We now call `setTimeout` and `setImmediate` only after we have accumulated 5ms of lack of sleep.
This solves the problem.

Issue #492, opened 2023-05-01
Panic in runtime_service


tomaka:
```
 108, 208, 22, 139, 191, 235, 131, 226, 1, 110, 7, 202, 77, 219, 57, 0, 127]), amount: 105642100001001 }), Ibc(Events { events: [Ok(UpdateClient { client_id: [49, 48, 45, 103, 114, 97, 110, 100, 112, 97, 45, 50, 56], client_type: [49, 48, 45, 103, 114, 97, 110, 100, 112, 97], revision_height: 1915715, revision_number: 2019, consensus_height: 2310752, consensus_revision_number: 2087 })] }), Balances(Deposit { who: AccountId32([109, 111, 100, 108, 80, 111, 116, 83, 116, 97, 107, 101, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), amount: 26410525000250 }), Balances(Deposit { who: AccountId32([109, 111, 100, 108, 108, 121, 97, 114, 116, 114, 115, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), amount: 79231575000751 })]
thread 'async-std/runtime' panicked at 'called `Option::unwrap()` on a `None` value', /home/dz/.cargo/git/checkouts/smoldot-2dcbf637e11a77d4/e4ebb26/light-base/src/runtime_service.rs:1403:118
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
After some hours of running, got 
```
                            match &mut guarded.tree {
                                GuardedInner::FinalizedBlockRuntimeKnown {
                                    tree, ..
                                } => {
                                    let idx = tree.input_iter_unordered().find(|block| block.user_data.hash == hash).unwrap().id; // 1403
                                    tree.input_set_best_block(idx);
                                }
                                GuardedInner::FinalizedBlockRuntimeUnknown { tree, .. } => {
                                    let idx = tree.input_iter_unordered().find(|block| block.user_data.hash == hash).unwrap().id;
                                    tree.input_set_best_block(idx);
                                }
                            }
```

```
smoldot-light = { git = "https://github.com/smol-dot/smoldot.git", default-features = false, features = ["std"]  }
smoldot = { git = "https://github.com/smol-dot/smoldot.git", default-features = false, features = ["std"] }
[[package]]
[...]

PR #497, opened 2023-05-02
`input_set_best_block` now properly accepts an `Option`

tomaka:
Fix https://github.com/smol-dot/smoldot/issues/492

I'm going to let the node run an hour or two to check if it doesn't panic.


tomaka:
Everything still working after ~1h45

It doesn't prove that the problem is fixed, but it's good enough for me

PR #505, opened 2023-05-03
Make wasm execution explicit and move CPU rate limit and periodic yield to the JS

tomaka:
This PR backports one of the changes of https://github.com/smol-dot/smoldot/pull/494

Instead of having `advance_execution` be called in all the callbacks called by the JS, we now have a separate JS task that is responsible only for executions. This JS task might get waken up by one of the callbacks.

Just like in #494, this allows moving the CPU rate limiting and periodic yield system to the JS code.

Contrary to #494, notifying that the execution is ready is done with a callback, and not by notifying a memory location. Callbacks are completely fine when things are single threaded, they just are inappropriate in case of multiple threads.

This PR de-spaghetifies the internals of smoldot. Everything is now in my opinion much more clear.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2534,6 +2534,7 @@ name = "smoldot-light-wasm"
 version = "1.0.3"
 dependencies = [
  "async-executor",
+ "async-task",
  "event-listener",
  "fnv",
  "futures-util",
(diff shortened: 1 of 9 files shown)


tomaka:
I've just realized a caveat: we no longer yield between two consecutive tasks executions.

PR #526, opened 2023-05-05
Merge `instance.ts` into `client.ts`

tomaka:
Continuation of the wasm-node JS code refactoring to make it other-thread-friendly.

We now merge `instance.ts` and `client.ts` into one, as the current split doesn't make much sense.
This also cleans up the code quite a lot, as warts have accumulated a bit over time.


tomaka:
CI fails because `start_shutdown` currently triggers a panic (intentionally). One piece of code that this PR removes is a flag that absorbs panics, which we set right before shutting down.
This will be properly fixed with https://github.com/smol-dot/smoldot/pull/527.

PR #528, opened 2023-05-05
Unify the three implementations of parseMultiaddr

tomaka:
Rather than have each platform implement `parseMultiaddr`, we now share the implementation. Each platform now tweaks the client options to forbid the connection types it doesn't support.


tomaka:
This is again a step towards instance-in-the-background, as otherwise we'd have to duplicate these functions again.

PR #552, opened 2023-05-09
Deploy tests coverage report on the gh pages

tomaka:
This PR generates HTML tests coverage information as described here: https://doc.rust-lang.org/rustc/instrument-coverage.html

It then deploys it onto the gh-pages for display.

This does **not** print a report on the PR about a change in the coverage or things like that, just a report about the `main` branch, which I think is more than enough.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -109,7 +109,7 @@ jobs:
         with:
           name: code-coverage-report
           path: ./upload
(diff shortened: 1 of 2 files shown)


tomaka:
Seems to work, but the insane time that `docs-publish` takes is a bit prohibitive.
Building the test coverage should be done in parallel of the regular docs building, and with a different cache.

PR #564, opened 2023-05-10
Fix the tests coverage again

### LINKED ISSUE #559: Tests coverage reports 0% for everything
https://smol-dot.github.io/smoldot/tests-coverage/index.html

I didn't actually notice that problem in https://github.com/smol-dot/smoldot/pull/552 because I assumed it worked, but actually no.
For what it's worth, locally I get non-zero values.
I suppose that something isn't configured properly.

tomaka:
Close #559


tomaka:
This time the report shows non-zero.

PR #603, opened 2023-05-25
Fix panic in case of invalid checkpoint in chain spec

### LINKED ISSUE #598: Panic in case of invalid checkpoint
See OP of https://github.com/paritytech/subxt/pull/965#issue-1715992252

tomaka:
Fix #598 

cc @lexnv

Would be great to have a test case, but generating one myself would legitimately take me several hours.
@lexnv If you could upload somewhere the chain spec JSON that triggers a panic so that I include it as a regression test in this PR, it would be great :pray: If not possible or too annoying, then it's not a big deal.


lexnv:
Hi,

This is the spec that I was using:
https://gist.github.com/lexnv/9050f533d8ec0bc5c31b897b007508da

### Before this patch

```bash
thread 'main' panicked at 'attempt to subtract with overflow', /Users/lexnv/.cargo/git/checkouts/smoldot-2dcbf637e11a77d4/43b3306/lib/src/chain_spec.rs:445:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### After this patch

```bash
Error: LightClient(AddChainError("Invalid checkpoint in chain specification: EmptyBabeEpochsList"))
```

lexnv (approved):

Issue #610, opened 2023-05-25
Wrong names for `transaction_unstable_watchEvent` events


tomaka:
See https://github.com/paritytech/subxt/pull/965/files#r1205387466

PR #611, opened 2023-05-25
Fix `author_extrinsicUpdate` notifications being PascalCased

tomaka:
Fix https://github.com/smol-dot/smoldot/issues/610


lib/src/json_rpc/methods.rs
```diff
@@ -1062,15 +1062,25 @@ pub enum SystemPeerRole {
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub enum TransactionStatus {
```
lexnv, line 1064: nit: `#[serde(rename_all = "camelCase")]` could also work here 🙏

tomaka: I'm aware of this, but I don't like the magic of serde deciding of the names.

PR #615, opened 2023-05-25
Publish v1.0.7

tomaka:


lexnv:
Amazing release and thanks for the help wrt the light-client integration in subxt!

Would it also be possible to publish smoldot and smoldot-light on crates.io? 

Thanks! 🙏

PR #631, opened 2023-05-29
Full node: move the storage handling and runtime handling to the consensus service

tomaka:
At the moment, the optimistic syncing strategy holds internally the diff between the finalized block and the best block, and holds internally the runtime of the finalized block and the best block, while the consensus service is only responsible for the storage of the finalized block.

While this makes a more convenient API, it is not a scalable solution as the diff between the finalized block and best block can in theory be huge. Instead, this diff should be stored (in some way or the other) in the database. This PR is a step in that direction by moving the handling of these diffs to the consensus service.

This is a preliminary step towards https://github.com/smol-dot/smoldot/issues/272, as for https://github.com/smol-dot/smoldot/issues/272 we will need to move the cache of Merkle value calculations outside of the `runtime_host`, and moving it to the syncing (where the entire trie isn't known) is way too complicated. The cache clearly must be owned by the consensus service.
This is also a step towards finishing the implementation of the `AllSync` for the full node, which is currently blocked because it is too complicated to handle these storage diffs in a tree fashion efficiently.

~~This PR is still a draft because, while everything works, the full node now OOMs. This is because the changes in the consensus service have been implemented in a naive way by duplicating the entire trie once per block. The Polkadot trie is around 70 MiB at the location where the syncing is on my machine, and because there's a justification only every 1024 blocks (I think?), one would need 70 GiB of RAM in order to sync, which I don't have.
Instead, the consensus service should now store the trie of non-finalized blocks in the database and then load values from the database.~~

### DIFF
--- a/full-node/src/run/consensus_service.rs
+++ b/full-node/src/run/consensus_service.rs
@@ -231,36 +231,36 @@ impl ConsensusService {
                 // the chain and the machine of the user.
                 NonZeroU32::new(2000).unwrap()
             },
-            full: Some(all::ConfigFull {
-                finalized_runtime: {
-                    // Builds the runtime of the finalized block.
-                    // Assumed to always be valid, otherwise the block wouldn't have been
-                    // saved in the database, hence the large number of unwraps here.
-                    let heap_pages = executor::storage_heap_pages_to_value(
-                        finalized_block_storage
-                            .node(bytes_to_nibbles(b":heappages".iter().copied()))
-                            .into_occupied()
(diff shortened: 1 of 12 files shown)


tomaka:
Pushed a commit that loads all the storage data from the database.

This breaks the database format. No migration system exists because the full node is experimental anyway.

The full node is significantly slower than before due to the many database accesses and the non-optimal way the storage is stored in the database. However there's no point in optimizing now, as we will need to make more changes to the database for #272.

PR #640, opened 2023-05-30
Use the browser native `DecompressionStream` API

tomaka:
Unfortunately, the DOM type definitions haven't been updated yet with the stabilization of `DecompressionStream`, so this PR doesn't compile. But eventually it will.


tomaka:
Tried in Firefox and Chrome. Works fine.

PR #665, opened 2023-06-04
Enable TCP NODELAY on Deno

tomaka:
Probably shouldn't be merged yet, even though the change is important, because Deno v1.29 isn't available on snap: https://snapcraft.io/deno

cc https://github.com/denoland/deno_install/issues/251


tomaka:
I'm going to merge this, as the chances that this `snap` package gets updated any time soon are slim. Plus I've noticed the `snap` issue because that's how I use it, but other distribution methods might have the problem as well without me being aware of it. In other words, this is not my problem.

PR #670, opened 2023-06-05
Properly handle requests for `NextKey` and `ClosestDescendantMerkleValue` when reading proofs

### LINKED ISSUE #272: RPC call returns "PrefixKeysForbidden" error code

tomaka:
Close #272

This finishes the plumbing.

### DIFF
--- a/lib/src/trie/proof_decode.rs
+++ b/lib/src/trie/proof_decode.rs
@@ -41,7 +41,7 @@
 use super::{nibble, trie_node, TrieEntryVersion};
 
 use alloc::{collections::BTreeMap, vec, vec::Vec};
-use core::{fmt, mem, ops};
+use core::{fmt, iter, mem, ops};
 
 /// Configuration to pass to [`decode_and_verify_proof`].
 pub struct Config<'a, I> {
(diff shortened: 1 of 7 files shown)


lib/src/trie/proof_decode.rs
```diff
@@ -658,7 +703,183 @@ impl<T: AsRef<[u8]>> DecodedTrieProof<T> {
...
+    /// value.
+    /// Returns `Some(None)` if the proof indicates that there is no descendant.
+    // TODO: return a Result instead of Option?
+    // TODO: accept params as iterators rather than slices?
+    pub fn closest_descendant_merkle_value(
+        &'_ self,
+        key: &[nibble::Nibble],
+    ) -> Option<Option<&'_ [u8]>> {
+        if key.is_empty() {
+            todo!()
```
tomaka, line 837: Yeah no

Issue #661, opened 2023-06-03
Full node is extremely slow


tomaka:
After https://github.com/smol-dot/smoldot/pull/639, the full node has become insanely slow, taking several minutes per block.

The reason for that is a very unoptimized database schema and the removal of caches.

I think that, just like Substrate, we should store the entire trie (including branch nodes) in the database. This would also include the Merkle value of each trie node.

Maybe the easiest way to achieve that is to have a `trie_node` table with `merke_value` as its primary key and columns like `child0_merkle_value`, `child0_partial_key`, `child1_merkle_value`, `child1_partial_key`, etc.

PR #715, opened 2023-06-09
Make `trie_root_calculator` generate events about trie changes

tomaka:
First step towards #661.

The `trie_root_calculator` now generates events indicating the Merkle values updates of the trie nodes, and which trie nodes (including branch nodes) have been destroyed.

The roadmap is to propagate these events throughout the stack so that the `full_node::consensus_service` writes these events into the database.


tomaka:
Full node still works.

PR #722, opened 2023-06-10
Add child trie unit tests for runtime_host

tomaka:
Close https://github.com/smol-dot/smoldot/issues/166


tomaka:
@pgherveou I've pushed a dummy commit with your email so that you are marked as co-author of this PR, is that ok?

pgherveou:
👍 will try to take a look at is as well

PR #777, opened 2023-06-19
No longer consider proof response with missing field as valid

tomaka:
This PR removes a work-around that was present due to an issue in Substrate.
This issue [was fixed](https://github.com/paritytech/substrate/pull/12732) seven months ago, and as such we can remove the work-around.


tomaka:
I'm not completely sure whether to merge this, as it's not impossible for a parachain team to use an old version of Substrate that doesn't have the fix.

PR #779, opened 2023-06-19
Do not send more than one request per peer at a time

tomaka:
This PR introduces a "request lock" system to the `NetworkService` of the light client.
When a "request lock" is held, no other request lock with the same PeerId can be created. A "request lock" can then be used to start an actual request.

This causes the `sync_service` (and all parts of the node) to only start one request per peer at a time.
This effectively implements an existing requirement of the Polkadot networking protocol that isn't enforced by Substrate.

Further PRs will further refine this system.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2418,6 +2418,7 @@ version = "0.6.0"
 dependencies = [
  "async-lock",
  "blake2-rfc",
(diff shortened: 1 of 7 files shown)


tomaka:
Just tried rebasing this PR, and I think that it's easier to restart it from scratch.

Issue #134, opened 2023-02-09
Add support for offchain workers


tomaka:
https://github.com/paritytech/smoldot/issues/73

tomaka:
The first step would be to implement all the offchain workers functions in `host.rs`, then add an `offchain_host.rs` file that executes the offchain worker of a runtime.

PR #788, opened 2023-06-20
offchain implementation

ermalkaleci:


tomaka:
Thanks for the PR!
cc #134 

While the changes in `host.rs` are mostly good, `runtime_host.rs` is dedicated to executing non-offchain functions. The runtime isn't allowed to call offchain worker functions, and so all these new variants should make `runtime_host.rs` return an error.

For executing the offchain workers, there should be a new module similar to `read_only_runtime_host.rs`.

lib/src/executor/offchain_runtime_host.rs
```diff
@@ -0,0 +1,536 @@
...
+                        }
+                    }
+                }
+
+                host::HostVm::ExternalStorageSet(req) => {
+                    let key = req.key().as_ref().to_vec();
+                    let value = req.value().map(|v| v.as_ref().to_vec());
+                    self.overlay_storage.insert(key, value);
+                    self.vm = req.resume()
+                }
```
tomaka, line 425: I'm pretty sure the offchain worker isn't supposed to modify the storage during the execution, no?
Does this block of code ever get reached?

ermalkaleci: yes it does. It will initialize the block and set current hash. technically it can clear storage as welll but should it be supported, I don't know

ermalkaleci: https://github.com/paritytech/substrate/blob/6bae274f42520c38499e0091ebc50b68cff00c4e/frame/executive/src/lib.rs#L660-L677

tomaka: After some chatting, indeed offchain workers can do anything, and we just have to discard the overlay changes at the end.
(I've also opened https://github.com/smol-dot/smoldot/issues/828)

Sorry for the extra work, but please revert https://github.com/smol-dot/smoldot/pull/788/commits/87a2c2e2f2542c84ca3fc021a0c0d82abb0b6264 then :pray: 

I think it makes sense for the offchain-worker-related stuff to be in `runtime_host` as well, with the intention that `runtime_host` becomes the unique way to call a runtime.

However, instead of having one variant per offchain-worker-related function, I think the offchain-worker-related functions should be grouped in another enum.

In other words, not like this:

```rust
enum RuntimeHostVm {
    StorageGet,
    NextKey,
    ...
    OffchainGet,
    OffchainSeed,
    ...
}
```

But like that:

```rust
enum RuntimeHostVm {
    StorageGet,
    NextKey,
    ...
    Offchain(OffchainContext),
    ...
}

enum OffchainContext {
    Get,
    Seed,
    SubmitTransaction,
    ...
}
```

This makes it easier for the API user of `runtime_host` to error if an offchain worker function is called by the runtime.

lib/src/verify/header_body.rs
```diff
@@ -592,6 +592,9 @@ impl VerifyInner {
                     self.inner = sig.verify_and_resume();
                     self.phase = phase;
                 }
+                (runtime_host::RuntimeHostVm::Offchain(_inner), _phase) => {
+                    unreachable!()
```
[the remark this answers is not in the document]
ermalkaleci: yeah they're not supposed to. updating...

tomaka:
I'll do a close look review later to double check that for example all the parameters have the good types, but other than that the approach is :+1:

lib/src/executor/runtime_host.rs
```diff
@@ -396,6 +396,7 @@ pub enum ErrorDetail {
     },
     /// Size of the logs generated by the runtime exceeds the limit.
     LogsTooLong,
+    ForbiddenHostCall,
```
tomaka, line 399: Please add one error in `validate.rs`, one in `header_body.rs`, etc. instead of here

ermalkaleci:
> I'll do a close look review later to double check that for example all the parameters have the good types, but other than that the approach is 👍

there're more offchain calls that needs to be implemented but we can address those in future PR

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
+    }
+}
+
+impl fmt::Debug for ExternalOffchainStorageGet {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_tuple("ExternalOffchainStorageGet").finish()
+    }
+}
+
+/// Must return the current timestamp.
```
tomaka, line 3101:
```suggestion
/// Must return the current UNIX timestamp.
```

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
 
-impl fmt::Debug for ExternalOffchainStorageSet {
+impl fmt::Debug for OffchainTimestamp {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_tuple("ExternalOffchainStorageSet").finish()
+        f.debug_tuple("OffchainTimestamp").finish()
+    }
+}
+
+/// Must return a random seed.
```
tomaka, line 3122:
```suggestion
/// Must provide a randomly-generate number.
```

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
+
+    /// Pointer to the transaction whose value must be set. Guaranteed to be in range.
+    tx_ptr: u32,
+
+    /// Size of the transaction whose value must be set. Guaranteed to be in range.
+    tx_size: u32,
+}
+
+impl OffchainSubmitTransaction {
+    pub fn transaction(&'_ self) -> impl AsRef<[u8]> + '_ {
```
tomaka, line 3165:
```suggestion
    /// Returns the SCALE-encoded transaction to submit to the chain.
    pub fn transaction(&'_ self) -> impl AsRef<[u8]> + '_ {
```

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
+
+        let value_len_enc = util::encode_scale_compact_usize(value_total_len);
+        self.inner.alloc_write_and_return_pointer_size(
+            host_fn.name(),
+            iter::once(&[1][..])
+                .chain(iter::once(value_len_enc.as_ref()))
+                .map(either::Left)
+                .chain(iter::once(value).map(either::Right)),
+        )
+    }
```
tomaka, line 3190: I think that's wrong: https://github.com/paritytech/substrate/blob/689da495a0c0c0c2466fe90a9ea187ce56760f2d/primitives/io/src/lib.rs#L1243

The return value is a `Result<(), ()>`, presumably indicating whether the transaction has successfully been submitted.

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
+impl OffchainSubmitTransaction {
+    pub fn transaction(&'_ self) -> impl AsRef<[u8]> + '_ {
+        self.inner
+            .vm
+            .read_memory(self.tx_ptr, self.tx_size)
+            .unwrap()
+    }
+
+    /// Resumes execution after having set the value.
+    pub fn resume(self, value: (&[u8], usize)) -> HostVm {
```
tomaka, line 3173:
```suggestion
    /// Resumes execution after having submitted the transaction.
    pub fn resume(self, success: bool) -> HostVm {
```

lib/src/executor/runtime_host.rs
```diff
@@ -412,6 +412,8 @@ pub enum RuntimeHostVm {
     NextKey(NextKey),
     /// Verifying whether a signature is correct is required in order to continue.
     SignatureVerification(SignatureVerification),
+    /// Offchain context is required in order to continue.
```
tomaka, line 415:
```suggestion
    /// Functions that can only be called within the context of an offchain worker.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+            }
+            // We only create a `OffchainStorageGet` if the state is one of the above.
+            _ => unreachable!(),
+        };
+
+        self.inner.run()
+    }
+}
+
+/// Loading timestamp is required in order to continue.
```
tomaka, line 942:
```suggestion
/// Providing the current UNIX timestamp is required in order to continue.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+#[must_use]
+pub struct OffchainTimestamp {
+    inner: Inner,
+}
+
+impl OffchainTimestamp {
+    /// Injects timestamp.
```
tomaka, line 949:
```suggestion
    /// Resume execution by providing the current UNIX timestamp.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+            }
+            // We only create a `OffchainTimestamp` if the state is one of the above.
+            _ => unreachable!(),
+        };
+
+        self.inner.run()
+    }
+}
+
+/// Loading random seed is required in order to continue.
```
tomaka, line 963:
```suggestion
/// Providing a random number is required in order to continue.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+#[must_use]
+pub struct OffchainRandomSeed {
+    inner: Inner,
+}
+
+impl OffchainRandomSeed {
+    /// Injects random seed.
```
tomaka, line 970:
```suggestion
    /// Resume execution by providing a random number.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+            }
+            // We only create a `OffchainRandomSeed` if the state is one of the above.
+            _ => unreachable!(),
+        };
+
+        self.inner.run()
+    }
+}
+
+/// Submit transaction is required in order to continue.
```
tomaka, line 984:
```suggestion
/// The runtime requests submitting a transaction.
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+#[must_use]
+pub struct OffchainSubmitTransaction {
+    inner: Inner,
+}
+
+impl OffchainSubmitTransaction {
+    pub fn transaction(&'_ self) -> impl AsRef<[u8]> + '_ {
```
tomaka, line 991:
```suggestion
    /// Returns the SCALE-encoded transaction that must be submitted.
    pub fn transaction(&'_ self) -> impl AsRef<[u8]> + '_ {
```

lib/src/executor/runtime_host.rs
```diff
@@ -884,6 +909,106 @@ impl SignatureVerification {
...
+        match &self.inner.vm {
+            host::HostVm::OffchainSubmitTransaction(req) => req.transaction(),
+            // We only create a `OffchainSubmitTransaction` if the state is one of the above.
+            _ => unreachable!(),
+        }
+    }
+    /// Injects outcome.
+    pub fn inject_outcome(mut self, value: impl AsRef<[u8]>) -> RuntimeHostVm {
```
tomaka, line 999:
```suggestion

    /// Resume execution. Must indicate whether the transaction has been successfully submitted.
    pub fn resume(mut self, success: bool) -> RuntimeHostVm {
```

lib/src/executor/runtime_host/tests.rs
```diff
@@ -179,6 +179,9 @@ fn execute_blocks() {
 
                     execution = req.inject_key(next_key.map(|nk| nk.into_iter()));
                 }
+                RuntimeHostVm::Offchain(_req) => {
+                    unimplemented!("Offchain calls are not supported yet.")
```
tomaka, line 183:
```suggestion
                    unimplemented!()
```

They will intentionally never be supported in this test.

lib/src/verify/header_body.rs
```diff
@@ -213,6 +213,8 @@ pub enum Error {
     /// Block has modified the `:heappages` key in a way that fails to parse.
     #[display(fmt = "Block has modified `:heappages` key in invalid way: {_0}")]
     HeapPagesParseError(executor::InvalidHeapPagesError),
+    /// Runtime called a forbidden host call.
```
tomaka, line 216:
```suggestion
    /// Runtime called a forbidden host function.
```

lib/src/transactions/validate.rs
```diff
@@ -213,6 +213,8 @@ pub enum Error {
     /// the runtime to always provide a non-empty list of tags. This error is consequently a bug
     /// in the runtime.
     EmptyProvidedTags,
+    /// Runtime called a forbidden host call.
```
tomaka, line 216:
```suggestion
    /// Runtime called a forbidden host function.
```

light-base/src/json_rpc_service/background.rs
```diff
@@ -1341,6 +1346,8 @@ enum RuntimeCallError {
         /// Version that the runtime supports.
         actual_version: u32,
     },
+    /// Runtime called a forbidden host call.
```
tomaka, line 1349:
```suggestion
    /// Runtime called a forbidden host function.
```

light-base/src/json_rpc_service/background/chain_head.rs
```diff
@@ -1281,6 +1281,16 @@ impl<TPlat: PlatformRef> ChainHeadFollowTask<TPlat> {
                                         runtime_host::RuntimeHostVm::SignatureVerification(sig) => {
                                             runtime_call = sig.verify_and_resume();
                                         }
+                                        runtime_host::RuntimeHostVm::Offchain(ctx) => {
+                                            runtime_call_lock.unlock(runtime_host::RuntimeHostVm::Offchain(ctx).into_prototype());
+                                            subscription.send_notification(methods::ServerToClient::chainHead_unstable_callEvent {
+                                                subscription: (&subscription_id).into(),
+                                                result: methods::ChainHeadCallEvent::Error {
+                                                    error: "Offchain calls are not supported".to_string().into(),
```
tomaka, line 1289:
```suggestion
                                                    error: "Runtime has called an offchain host function".into(),
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
-            HostFunction::ext_offchain_submit_transaction_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_network_state_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_timestamp_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_sleep_until_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_random_seed_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_local_storage_set_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_local_storage_compare_and_set_version_1 => {
+            HostFunction::ext_offchain_is_validator_version_1 => HostVm::ReadyToRun(ReadyToRun {
+                inner: self.inner,
+                resume_value: Some(vm::WasmValue::I32(1)),
```
tomaka, line 1728:
```suggestion
                resume_value: Some(vm::WasmValue::I32(1)),   // TODO: ask the API user
```

I'm fine with partially-implemented stuff, but I would like TODOs to indicate where things are not finished.

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
-            HostFunction::ext_offchain_local_storage_get_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_offchain_local_storage_clear_version_1 => host_fn_not_implemented!(),
+            HostFunction::ext_offchain_random_seed_version_1 => {
+                HostVm::OffchainRandomSeed(OffchainRandomSeed {
+                    inner: self.inner,
+                    calling: id,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_set_version_1 => {
+                let _kind = expect_u32!(0);
```
tomaka, line 1755:
```suggestion
                let _kind = expect_u32!(0);  // TODO: parse and provide the storage kind
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
+                    key_ptr,
+                    key_size,
+                    value: Some((value_ptr, value_size)),
+                    old_value: None,
+                    inner: self.inner,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_compare_and_set_version_1 => {
+                let _kind = expect_u32!(0);
```
tomaka, line 1767:
```suggestion
                let _kind = expect_u32!(0);  // TODO: parse and provide the storage kind
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
+                    key_ptr,
+                    key_size,
+                    value: Some((value_ptr, value_size)),
+                    old_value: Some((old_value_ptr, old_value_size)),
+                    inner: self.inner,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_get_version_1 => {
+                let _kind = expect_u32!(0);
```
tomaka, line 1780:
```suggestion
                let _kind = expect_u32!(0);  // TODO: parse and provide the storage kind
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+
+                HostVm::ExternalOffchainStorageGet(ExternalOffchainStorageGet {
+                    key_ptr,
+                    key_size,
+                    calling: id,
+                    inner: self.inner,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_clear_version_1 => {
+                let _kind = expect_u32!(0);
```
tomaka, line 1791:
```suggestion
                let _kind = expect_u32!(0);  // TODO: parse and provide the storage kind
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+                HostVm::OffchainRandomSeed(OffchainRandomSeed {
+                    inner: self.inner,
+                    calling: id,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_set_version_1 => {
+                let _kind = expect_u32!(0);
+                let (key_ptr, key_size) = expect_pointer_size_raw!(1);
+                let (value_ptr, value_size) = expect_pointer_size_raw!(2);
+                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```
tomaka, line 1758:
```suggestion
                // TODO: the API user should be able to differentiate between the `ext_offchain_index_*` and `ext_offchain_*` functions
                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+                    old_value: None,
+                    inner: self.inner,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_compare_and_set_version_1 => {
+                let _kind = expect_u32!(0);
+                let (key_ptr, key_size) = expect_pointer_size_raw!(1);
+                let (old_value_ptr, old_value_size) = expect_pointer_size_raw!(2);
+                let (value_ptr, value_size) = expect_pointer_size_raw!(3);
+                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```
tomaka, line 1771:
```suggestion
                // TODO: the API user should be able to differentiate between the `ext_offchain_index_*` and `ext_offchain_*` functions
                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```

lib/src/executor/host.rs
```diff
@@ -1702,21 +1719,85 @@ impl ReadyToRun {
...
+                    key_ptr,
+                    key_size,
+                    calling: id,
+                    inner: self.inner,
+                })
+            }
+            HostFunction::ext_offchain_local_storage_clear_version_1 => {
+                let _kind = expect_u32!(0);
+                let (key_ptr, key_size) = expect_pointer_size_raw!(1);
+                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```
tomaka, line 1793:
```suggestion
                // TODO: the API user should be able to differentiate between the `ext_offchain_index_*` and `ext_offchain_*` functions
                HostVm::ExternalOffchainStorageSet(ExternalOffchainStorageSet {
```

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
         }
     }
 
+    /// Returns the old value.
```
tomaka, line 3017:
```suggestion
    /// Returns the value the current value should be compared against. The operation is a no-op if they don't compare equal.
```

lib/src/executor/host.rs
```diff
@@ -2930,18 +3014,185 @@ impl ExternalOffchainStorageSet {
...
+    pub fn old_value(&'_ self) -> Option<impl AsRef<[u8]> + '_> {
+        if let Some((ptr, size)) = self.old_value {
+            Some(self.inner.vm.read_memory(ptr, size).unwrap())
+        } else {
+            None
+        }
+    }
+
     /// Resumes execution after having set the value.
```
tomaka, line 3026:
```suggestion
    /// Resumes execution after having set the value. Must indicate whether a value was written.
```

lib/src/executor/runtime_host.rs
```diff
@@ -1310,12 +1435,39 @@ impl Inner {
...
+                    };
+
+                    if replace {
+                        if let Some(value) = req.value() {
+                            self.offchain_storage_changes
+                                .insert(req.key().as_ref().to_vec(), Some(value.as_ref().to_vec()));
+                        } else {
+                            self.offchain_storage_changes.remove(req.key().as_ref());
+                        }
+                    }
```
tomaka, line 1468: That's not correct unfortunately for the compare-and-set variant.
If there's no existing value in the overlay, we have to query the current value from the API user in order to compare it with what `old_value` returns.

Please either implement properly or remove support for the `compare_and_set` function. :pray:

tomaka (changes requested):
Lots of nits, but the only "big" change is the fact that compare-and-set is misimplemented

lib/src/executor/runtime_host.rs
ermalkaleci: oh that's right. We need `compare_and_set`, so all I need is insert value to `offchain_storage_changes` when it's injected to `OffchainStorageGet`

tomaka: Maybe I'm misunderstanding, but to me `compare_and_set` means: load the current value (with an `OffchainStorageGet` if necessary), compare with the value that the runtime provides, and only if they're different then write the new value that the runtime provides.

ermalkaleci: Actually naming `overlay` is not correct (updating) because offchain worker doesn't have overlay, just read/write from storage.
The `old_value` provided by runtime is read using `OffchainStorageGet` first which is inserted into `offchain_storage_chages` https://github.com/smol-dot/smoldot/pull/788/files#diff-3cdf7d07bd6351e0742505e7dc4e11464f5c12424357a082ae6b3249d8b28298R932-R935

tomaka: Sorry for the delay.

After looking into Substrate, the `ext_offchain_index` functions are affected by the storage transactions: https://github.com/paritytech/substrate/blob/689da495a0c0c0c2466fe90a9ea187ce56760f2d/primitives/state-machine/src/overlayed_changes/mod.rs#L790-L843

However, the `ext_offchain_local_storage` functions indeed don't have an overlay and just directly load and dump from the database.

ermalkaleci: I didn't check `ext_offchain_index`, I thought it was the same as `local_storage`. I will have another look

ermalkaleci: ok so `ext_offchain_index_set/clear` is a way for on-chain to modify off-chain storage if transaction is successful and that's why this writes to overlay because it is committed or reverted the same as transaction
https://github.com/paritytech/substrate/blob/7b1d6f337fd1cee4e1fe675c89c8d82ce56d5c47/frame/session/src/historical/offchain.rs#L215

ermalkaleci: @tomaka I suggest we update `ext_offchain_index_set/clear` methods to skip and continue execution and address this later.

tomaka: I don't think it's very complicated to implement properly:

- In `host.rs`, add back a variant for `ext_offchain_index_set/clear`, not the same as the one for `ext_offchain_local_storage_*` so that `runtime_host.rs` can know the difference.
- In `runtime_host.rs`, remove `offchain_storage_changes` from `Config` and `Success`.
- Add a `OffchainStorageSet` variant in `runtime_host`.
- Add a field `offchain_storage_changes` to `PendingStorageChanges` (`PendingStorageChanges` stores the state during a transaction).
- Around [here](https://github.com/smol-dot/smoldot/blob/7f6a4662a25b482f2fe8b991c291a39bfe31dbef/lib/src/executor/runtime_host.rs#L1088), add a `if matches!(self.vm, HostVm::Finished) && !self.offchain_storage_changes.is_empty() { ... }`, and in the `if` you remove an element from `offchain_storage_changes` and yield a `OffchainStorageSet`. This commits the overlay offchain storage changes.

ermalkaleci: I added `offchain_index` into `PendingStorageChanges` and they're applied into `offchain_storage_changes` on success.

tomaka:
I thought it was easier to just do the change I wanted by writing code than explain it in English, so I've just pushed to your branches some commits that finish the changes.

Please let me know what you think.

tomaka (approved):

ermalkaleci:
> I thought it was easier to just do the change I wanted by writing code than explain it in English, so I've just pushed to your branches some commits that finish the changes.
> 
> Please let me know what you think.

hey, thank you for the update. I like this implementation more

lib/src/executor/runtime_host.rs
Wsteth: Hi there,
I found a problem when the host executes ExternalOffchainIndexSet.
https://github.com/smol-dot/smoldot/blob/37482d9a635a17f9e6a77aa245b4684b2f72e750/lib/src/executor/runtime_call.rs#L1750-L1759.

The offchain storage will not be changed, and when I use offchainStorageGet, it doesn't work.
It seems there is nothing done with add pending_storage_changes, and when I remove pending_storage_changes, I can get the storage value.

Issue #798, opened 2023-06-23
Backport JSON-RPC storage multiple items


tomaka:
https://github.com/paritytech/json-rpc-interface-spec/pull/51

Issue #605, opened 2023-05-25
Backport changes to `chainHead_unstable_storage`


tomaka:
cc https://github.com/paritytech/json-rpc-interface-spec/pull/37

PR #806, opened 2023-06-23
Simplify chain-head-related function implementations

tomaka:
cc https://github.com/smol-dot/smoldot/issues/605 https://github.com/smol-dot/smoldot/issues/798

This PR significantly simplifies the implementations of the various `chainHead` functions, notably by removing duplicate information and using early returns.

This is now possible thanks to https://github.com/smol-dot/smoldot/pull/793.

I'm still not happy with the complexity of the main implementation of `chainHead_follow`, but this PR is a big enough step that it's worth merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2435,6 +2435,7 @@ dependencies = [
  "event-listener",
  "fnv",
  "futures-channel",
+ "futures-lite",
  "futures-util",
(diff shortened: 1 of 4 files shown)


tomaka:
`futures_lite::race` requires `std` because they use randomness to poll the futures :face_exhaling:

PR #820, opened 2023-06-25
fix: types for `"modeResolution": "Node"` TS projects

### LINKED ISSUE #787: typings on sub-path imports are broken
Basically, this doesn't work:

```js
import { run } from "smoldot/worker"
import { compileBytecode } from "smoldot/bytecode"
```

while this does work:

```js
import { run } from "smoldot/dist/mjs/worker-browser"
import { compileBytecode } from "smoldot/dist/mjs/bytecode-browser"
```

## [DEMO](https://stackblitz.com/edit/stackblitz-starters-9iolhk?file=package.json,src%2Fmain.ts)

josepot:
Fixes #787

### DIFF
--- a/wasm-node/javascript/.gitignore
+++ b/wasm-node/javascript/.gitignore
@@ -1,2 +1,5 @@
 node_modules
 dist
+worker
+no-auto-bytecode
+bytecode
--- a/wasm-node/javascript/create-subpath-for-legacy-types.sh
+++ b/wasm-node/javascript/create-subpath-for-legacy-types.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+#   https://github.com/smol-dot/smoldot/issues/787#issuecomment-1606157019
+#
+
+for SUBPATH in bytecode no-auto-bytecode worker
+do
+  mkdir -p $SUBPATH
+  cp ./dist/mjs/public-types.d.ts ./$SUBPATH/
+  cp ./dist/mjs/$SUBPATH-browser.d.ts ./$SUBPATH/index.d.ts
+done
--- a/wasm-node/javascript/package.json
+++ b/wasm-node/javascript/package.json
@@ -16,7 +16,10 @@
     "url": "https://github.com/smol-dot/smoldot/issues"
   },
   "files": [
-    "dist"
+    "dist",
+    "worker",
+    "no-auto-bytecode",
+    "bytecode"
   ],
   "main": "dist/cjs/index-browser.js",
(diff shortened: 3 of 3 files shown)


tomaka:
Did you check whether this solves the problem?

josepot:
> Did you check whether this solves the problem?

Yes, of course! What I tried is:

On `smoldot` I ran `git clean -dfx` while being in the branch of this PR. Then, inside the `wasm/javascript`  folder I ran `npm i && npm run build`.  Next, in the project where I was testing this issue I removed the `smoldot` folder inside the `node_modules` and I created it again, but empty. Finally, I copied inside that folder the `package.json` file from `smoldot`, and the following folders (which get created by the `npm run build` command): `dist`, `bytecode`, `no-auto-bytecode` and `worker`.

In the problematic TS project: I tried changing the `modeResolution` to `Node` and everything was working correctly, and then I switched it to `Node16` and everything was still working nicely.

So, yeah. I'm pretty sure that this should solve the problem.

tomaka:
But does your example project work with no `moduleResolution` configured? That's what I care the most about.

josepot:
> But does your example project work with no `moduleResolution` configured? That's what I care the most about.

If you don't setup the `moduleResolution` it works as long as you setup the `module` property to `CommonJS`. It also works if you setup `moduleResolution` to either `Node` or `Node16` or `NodeNext`, regardless of what you set the `module` property to.

It doesn't work if you setup `moduleResolution` to `classic`, or if you setup the `module` property to anything else other than `CommonJS` without setting up the `moduleResolution` property. However, that is to be expected and TS in the error tells you to set the `moduleResolution`. No other modern library works with typescript (out of the box) with `moduleResolution` set to `classic` or leaving `moduleResolution` unset if the `module` property is different than `CommonJS`.

In other words: I'm pretty sure that this is as good as it gets.

tomaka:
This PR as it is doesn't really make sense to me.

I don't care about people who use TypeScript < 4.7. What I would like is for things to work with `moduleResolution: classic`.

Either there is a way to make it work when `moduleResolution` is `classic` and we should apply it (if it's reasonable), or there is no way and we don't change anything.

tomaka:
[Given that TypeScript themselves recommend `moduleResolution: node`](https://www.typescriptlang.org/docs/handbook/module-resolution.html#module-resolution-strategies), I'm also relatively happy with forcing people to use `moduleResolution: node`, provided it's properly documented.

It's really hard to figure out what is in the mind of the designers of TypeScript. There's no consistency between the documentation and the default values of the options.

PR #836, opened 2023-06-30
Document using `moduleResolution: node`

### LINKED ISSUE #787: typings on sub-path imports are broken

tomaka:
Close #820 
Close #787

After some IRL discussions, this is IMO the best course of action.

cc @josepot

### DIFF
--- a/wasm-node/javascript/README.md
+++ b/wasm-node/javascript/README.md
@@ -125,3 +125,8 @@ import { compileBytecode } from 'smoldot/bytecode';
 compileBytecode().then((bytecode) => postMessage(bytecode))
 onmessage = (msg) => smoldot.run(msg.data);
 ```
+
+Note that importing sub-paths (for example importing `smoldot/worker`) relies on a relatively
+modern JavaScript feature. If you import a smoldot sub-path from a TypeScript file, you might have
+to configure TypeScript to use `"moduleResolution": "node"`. [The official TypeScript documentation
+itself recommends setting this configuration option](https://www.typescriptlang.org/docs/handbook/module-resolution.html#module-resolution-strategies).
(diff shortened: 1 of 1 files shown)


wasm-node/javascript/README.md
```diff
@@ -125,3 +125,8 @@ import { compileBytecode } from 'smoldot/bytecode';
 compileBytecode().then((bytecode) => postMessage(bytecode))
 onmessage = (msg) => smoldot.run(msg.data);
 ```
+
+Note that importing sub-paths (for example importing `smoldot/worker`) relies on a relatively
+modern JavaScript feature. If you import a smoldot sub-path from a TypeScript file, you might have
+to configure TypeScript to use `"moduleResolution": "node"`. [The official TypeScript documentation
```
josepot, line 131: I'm afraid that we must have misunderstood each other at some point 😬 , sorry about that.

The thing is that -as of right now- setting `"moduleResolution": "Node"` won't make the subpaths imports work. For instance, the demo that I setup when I opened #787 to demonstrate the issue with subpath typings [uses `"moduleResolution": "Node"` ](https://stackblitz.com/edit/stackblitz-starters-tuwp6y?file=package.json,tsconfig.json%3AL17). When I closed #787 I mentioned that [the subpaths typings work when setting `"moduleResolution": "Node16"`](https://github.com/smol-dot/smoldot/issues/787#issuecomment-1605737157).

The purpose of #820 is precisely to make the typings work when setting the `moduleResolution` field to either Node or Node16.

What I understood from the IRL discussion was that you were ok with the fact that TS users had to either use `"moduleResolution": "Node16"` if they wanted to use proper subpath imports, and require users that can't (or won't) use `"moduleResolution": "Node16"` to import the assets from the root path, rather than using subpath imports.

josepot (approved):

Issue #92, opened 2023-02-09
Add finalized block runtime code and heappages to wasm node database


tomaka:
https://github.com/paritytech/smoldot/issues/1769

tomaka:
Requires https://github.com/smol-dot/smoldot/issues/177

tomaka:
Can be done without #177 by storing the Merkle value of `:code` rather than the hash of its value.

tomaka:
This requires not just the Merkle value of `:code` but also the key of its closest ancestor, because we need to query specifically this closest ancestor at the follow-up warp sync.
Querying `:cod` for example will lead to `:code` being included in the proof in order to prove that `:cod` doesn't exist.

Issue #177, opened 2023-02-13
Add a network protocol to optimize state_getKeys


tomaka:
Instead of using the current method of asking for multiple proofs, we should be able to ask a full node for all the keys under a certain prefix in one shot.

Requires a Polkadot RFC

tomaka:
Opened https://github.com/w3f/PPPs/pull/5

Substrate implementation would be https://github.com/paritytech/substrate/pull/12843

skunert:
If I read the PPP correctly, resolving this issue would get rid of the "PrefixKeysForbidden" error code on some runtime calls, correct?

tomaka:
No, this is unrelated to #272

tomaka:
cc https://forum.polkadot.network/t/state-trie-migration/852/13

PR #863, opened 2023-07-06
Store the runtime code in the light client database

tomaka:
Close https://github.com/smol-dot/smoldot/issues/92

As explained in #92, we can't just store the runtime code in the database (doing so would require https://github.com/smol-dot/smoldot/issues/177), but the Merkle value of the trie node and the closest ancestor.

When the database is loaded back, the warp syncing queries said closest ancestor and compares the Merkle value of the relevant children with the one in the database.

In practice, the storage query is thus now only around 850 bytes, compared to ~1.0 to 1.5 MiB beforehand.
While this "only" saves around 300ms on my machine, I can imagine that people with not-so-good Internet connections will be happy.

I'm not super-super happy with the changes in the code, and I've added all these fields while having in mind that they will be removed again after #177, but they're also not *too* bad.

### DIFF
--- a/lib/src/sync/warp_sync.rs
+++ b/lib/src/sync/warp_sync.rs
@@ -112,6 +112,7 @@ use alloc::{
 };
 use core::{iter, mem, ops};
 
+pub use trie::Nibble;
 pub use verifier::{Error as FragmentError, WarpSyncFragment};
 
 mod verifier;
@@ -175,6 +176,9 @@ pub struct ConfigCodeTrieNodeHint {
 
     /// Storage value corresponding to [`ConfigCodeTrieNodeHint::merkle_value`].
     pub storage_value: Vec<u8>,
+
+    /// Closest ancestor of the `:code` key except for `:code` itself.
+    pub closest_ancestor_excluding: Vec<Nibble>,
 }
 
 /// Initializes the warp sync state machine.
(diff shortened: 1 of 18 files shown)


lib/src/sync/warp_sync.rs
```diff
@@ -620,9 +659,24 @@ impl<TSrc, TRq> InProgressWarpSync<TSrc, TRq> {
         let runtime_parameters_get = if let Phase::RuntimeDownload {
             header,
             warp_sync_source_id,
+            hint_doesnt_match,
+            downloaded_runtime: None,
```
tomaka, line 663: This specific line fixes a bug mentioned in the CHANGELOG where the warp syncing downloads the runtime code twice.

pgolovkin:
Is this caching works if I run the light client using the substrate-connect library?

tomaka:
That's more a question for substrate-connect, but yes for the 4 well-known chains (once substrate-connect has upgraded their smoldot of course).
The database is shared through the substrate-connect extension, so if you connect to Polkadot on one website, then connect to Polkadot on a different website, it will use what it has downloaded on the first one.

pgolovkin:
Do I understand correct that even if I don't have substrate-connect extension (I'm using the Electron environment) the latest state of the smoldot will be saved and when I run my Electron application next time this saved state will be used?

**The example:**
Given that:
1. Electron application with the substrate-connect
2. Polkadot latest chainspec in the smoldot (for example) with the latest block number 16342124

User actions:
1. The user runs the Electron application
2. Light client is started and synching
3. The user quits the Electron application, the latest synched block is 16342301
4. In a few hours the user starts the Electron application again

The question:
After starting the application (item 4) the Light Client will be synching from the block 16342301?

tomaka:
Substrate-connect is kind of pointless if you don't have the extension, and I would suggest to directly use smoldot.

PR #870, opened 2023-07-08
Rewrite Noise implementation

tomaka:
cc https://github.com/smol-dot/smoldot/issues/133

This PR switches from `snow` to directly using cryptographic primitives for the Noise protocol.

As you can see, this "only" adds 735 lines of code to `noise.rs`, which to me shows that the 6000 lines of code in the `snow` library are a bit exaggerated.

The benefits are:

- We get rid of `snow` that is unmaintained and dependencies are more up-to-date.
- It's a step towards #133.
- We can finally fix the fact that every single networking message gets memcpy'ed 2 or 3 times due to API issues.
- We can finally fix the fact that the connection state machine might stall if the outgoing buffer is too small, which fortunately isn't an issue in practice but has always bothered me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,41 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "aead"
-version = "0.4.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
-dependencies = [
- "generic-array",
-]
-
-[[package]]
-name = "aes"
-version = "0.7.5"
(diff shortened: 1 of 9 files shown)


tomaka:
PRs like these really should be reviewed, but I know that nobody is going to review it anyway, so there's no point in waiting to merge.

olanod:
> PRs like these really should be reviewed, but I know that nobody is going to review it anyway, so there's no point in waiting to merge.

Sounds sad, how could this be improved in the future? growing your team with more resources? would it help to place some kind of financial incentive on PRs to be reviewed(making incentives part of your budget) and share it with a competent group of people you know or more openly in a group like the fellowship?

Btw excited about `no_std` support :)

tomaka:
> Sounds sad, how could this be improved in the future? growing your team with more resources? would it help to place some kind of financial incentive on PRs to be reviewed(making incentives part of your budget) and share it with a competent group of people you know or more openly in a group like the fellowship?

The problem is that there's "review" and "review". It's easy to skim 5 minutes over the code and say "looks good to me".
For example, reviewing this PR requires knowledge not only of the smoldot networking but also of the Noise protocol. A proper review would take 10-ish hours.

I don't really have a solution here.

Issue #735, opened 2023-06-14
Remove the `JsonRpcResponses` type


tomaka:
I'm not very happy with the [JsonRpcResponses](https://docs.rs/smoldot-light/latest/smoldot_light/struct.JsonRpcResponses.html#) type in the public API of the light client.

I think that it should instead be a `pub fn next_json_rpc_response(&self, ChainId)` function of the `Client`.

This probably requires deeper changes to the API of the `Client`.

tomaka:
This issue is more or less tied to #91 

Doing this change would require changing the `Client` to have interior mutability.

At the moment, the `Client`'s API accepts `&mut self` everywhere, meaning that the Wasm node can do things such as adding a chain without running the risk of having to wait to lock a mutex.

The good news is that the JS API is ready to accomodate for an asynchronous `add_chain` function. Only the internal plumbing would need to be updated.

PR #931, opened 2023-07-18
Print average CPU usage in the wasm node periodic log line

tomaka:
This PR prints the average CPU usage in the log line that the wasm node periodically prints.

Example:

> [11:29:03.528] [smoldot] Smoldot v1.0.13. Current memory usage: 82.1 MiB. Average download: 1.8 MiB/s. Average upload: 17.5 kiB/s. Average CPU cores: 0.40.

I'm opening as draft because this doesn't take into account the time spent in `add_chain`, which is pretty significant.

In order to do this properly, we first need https://github.com/smol-dot/smoldot/issues/735


tomaka:
Oops, many commits here are not intended.

Issue #919, opened 2023-07-16
Noise and Yamux should use `read_write` system


tomaka:
It would considerably simplify the code of `established`.

Blocked by https://github.com/smol-dot/smoldot/issues/918

PR #943, opened 2023-07-21
Make `Noise` produce a `ReadWrite` of the decrypted stream

tomaka:
cc https://github.com/smol-dot/smoldot/issues/919


tomaka:
Tests are passing, but all requests time out when using the node.

I'll put this PR on the side and instead move to cleaning up the `ReadWrite`-related code in `full-node` and `light-base`.

PR #958, opened 2023-07-26
Integrate `Yamux` into the read_write system

### LINKED ISSUE #919: Noise and Yamux should use `read_write` system

tomaka:
Close #919


tomaka:
Beyond the failing tests which need to be either updated or removed, the networking unfortunately seems to have more trouble finding peers than before this PR and I'm not sure why.

tomaka:
> Beyond the failing tests which need to be either updated or removed, the networking unfortunately seems to have more trouble finding peers than before this PR and I'm not sure why.

Several tests in `established.rs` are failing, and fixing them will most likely fix the problem.

tomaka:
Things seem to work well after the latest two commits.

PR #1015, opened 2023-08-10
Calculate correct start slot for skipped epoch transition

skunert:
Small followup to https://github.com/smol-dot/smoldot/pull/991
- Removes some assertions that no longer hold now that epochs can be skipped
- When an epoch is skipped we need to adjust the starting slot of the next epoch


tomaka (approved):
It seems good to me, although to honest Babe now feels over-engineered and has become hard to reason about.

lexnv (approved):

Issue #864, opened 2023-07-06
Improve light client time to warp sync finished


tomaka:
Here's an example break down of how long it takes from smoldot's initialization (when we enter the smoldot Rust code) to warp syncing being completely finished in a best case scenario: (done on top of https://github.com/smol-dot/smoldot/pull/863)

- Parsing the chain spec (assuming a stripped down chain spec): 30ms
- Open first connection + various handshakes (WebSocket + encryption): 158ms
- Peering with that node: 56ms
- Download warp syncing proof: 85ms (in this example the proof is one fragment)
- Verify warp syncing proof: 82ms
- Download the hash of the runtime code to check whether it matches the one in cache: 56ms
- Build the runtime with wasmi: 328ms
- Download call proofs containing various storage items such as the current list of authorities: 60ms
- Perform some runtime calls to verify what we've just downloaded: 8ms

Overall that's roughly 415ms of waiting for networking messages, and 410ms of verifying signatures and compiling the runtime (where I rely on third party libraries that I don't really want to dig into).

What we can do is:

- Download the runtime and call proofs as soon as we have downloaded the end of the warp sync proof. We can simply assume that the proof is correct, and later discard the download if it turns out that it isn't. The call proofs can be done immediately before downloading the runtime, by assuming that the runtime is using the latest versions of all APIs. That would save 116ms in the example above.

- WebRTC should remove one round-trip time I believe, which would save 50 to 60ms.

- Look into why parsing the chain spec takes so long.

tomaka:
cc https://github.com/paritytech/wasmi/issues/732

tomaka:
With all the changes that have been done since this issue has been opened, I've just measured a 547ms initialization time with Westend.

Here are the logs:

```
[14:17:47.469] [smoldot] Smoldot v2.0.15
[14:17:47.484] [smoldot] Chain initialization complete for westend2. Name: "Westend". Genesis hash: 0xe143…423e. Database starting at: 0x1254…62f8 (#18797921)
[14:17:47.484] [smoldot] Chain specification of westend2 contains a non-empty `telemetryEndpoints` field. Smoldot doesn't support telemetry endpoints and as such this field is unused.
[14:17:47.484] [smoldot] Chain specification of westend2 contains a list of bad blocks. Bad blocks are not implemented in the light client. An appropriate way to silence this warning is to remove the bad blocks from the chain specification, which can safely be done:
[...]

PR #1060, opened 2023-08-23
Warp sync state machine refactoring

### LINKED ISSUE #67: Properly handle warp syncing failures
https://github.com/paritytech/smoldot/issues/2529

### LINKED ISSUE #119: GrandPa warp sync robustness to empty responses
https://github.com/paritytech/smoldot/issues/722

tomaka:
Close #67 
Close #119
cc https://github.com/smol-dot/smoldot/issues/864


tomaka:
It's unclear to me how to manage malicious sources the correct way.
Presently we do it with a very naive banning system where one error results in a ban, and all bans are reset if all sources are banned.

tomaka:
The difficulty in designing the warp sync state machine comes from https://github.com/smol-dot/smoldot/issues/1109

In this PR, I'm going to remove the concept of "successful warp sync" from the warp sync state machine, and instead modify the `AllSync` to treat the first successful warp sync as a trigger to switch to "all forks syncing". This way, https://github.com/smol-dot/smoldot/issues/1109 consists only in modifying `AllSync`.

tomaka:
> It's unclear to me how to manage malicious sources the correct way.

I think the best approach is to first query only the source(s) with the highest finalized block, but as requests fail broaden the search to all sources.

The high-level code must also be involved as it should remove sources that don't answer properly.

PR #1069, opened 2023-08-25
Install Rust the same way in all CI steps

tomaka:
Right now, some jobs use `container: image: rust:1.71` while some others use `actions-rs/toolchain`.
This PR switches everything to the `rust:1.71` image.

Fixes the CI failure in https://github.com/smol-dot/smoldot/pull/1065 and https://github.com/smol-dot/smoldot/pull/1068 by switching back to Rust 1.71 instead of the newly-released 1.72.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,6 +191,7 @@ jobs:
       image: rust:1.71
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


tomaka:
I gave up trying to upgrade `deno-publish`, as the error is beyond my understanding.

Issue #1006, opened 2023-08-09
Implement the new JSON-RPC and the most-commonly called legacy JSON-RPC functions in the full node


tomaka:

tomaka:
As highlighted in https://github.com/smol-dot/smoldot/pull/1073, it's not completely clear to me how to handle pruning.

When a block gets finalized, the `chainHead` API requires that the pruned blocks that were still pinned remain accessible (including their storage), while currently the database just removes them immediately.

The database code should be modified to have a more powerful pruning system rather than just "remove any block not a descendant of the finalized block".

One possibility could be for the full node to lie to the database and delay finalization by for example 32 blocks. However I think it's a bad idea to add a layer between the actual database and the rest of the code, as it would make it impossible to read the database without starting the node.

PR #1073, opened 2023-08-25
Implement `chain_subscribeAllHeads` for the full node

tomaka:
cc https://github.com/smol-dot/smoldot/issues/1006

This is a big leap, as it adds a blocks notifications system to the `consensus_service`.

It's unfortunately not clear to me yet how this will interact with the database pruning blocks. The idea of a consensus service running in the background and populating a separate database is actually not great IMO, and merging the database and consensus service seems like a good idea, but this is for a different PR.


tomaka:
> It's unfortunately not clear to me yet how this will interact with the database pruning blocks.

I'm thinking of holding off this PR before this is figured out, as otherwise I'll have to refactor the code anyway.

PR #1074, opened 2023-08-25
Implement `system_chain` and `system_properties` for the full node

tomaka:
cc https://github.com/smol-dot/smoldot/issues/1006


tomaka:
Let's merge https://github.com/smol-dot/smoldot/pull/1075 first and add some tests.

PR #1102, opened 2023-09-04
Implement `chainHead_follow` on the full node

tomaka:
Only implements following blocks, and not runtime calls or storage retrievals or header/body retrievals.

~~Before proceeding, we must figure out how to access the runtime from the JSON-RPC service.~~

~~Implementing the legacy API first might actually be a better idea, as it's easier.~~

### DIFF
--- a/full-node/src/consensus_service.rs
+++ b/full-node/src/consensus_service.rs
@@ -133,7 +133,7 @@ enum ToBackground {
     SubscribeAll {
         buffer_size: usize,
         // TODO: unused field
-        _max_pinned_blocks: NonZeroUsize,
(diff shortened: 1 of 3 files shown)


tomaka:
This PR gives to the per-client task the responsibility to spawn the task associated to `chainHead_follow`.
I think it might be a better idea to share this by spawning a task for `chainHead_follow` at initialization, while the per-client task would only enforce the limit to the number of follows.

tomaka:
> I think it might be a better idea to share this by spawning a task for chainHead_follow at initialization, while the per-client task would only enforce the limit to the number of follows.

I've changed my mind about this, due to the fact that closing subscriptions due to JSON-RPC client disconnects would be harder to track.

PR #1104, opened 2023-09-04
Turn `HostVmPrototype::registered_functions` into an `Arc` for faster cloning

tomaka:


lib/src/executor/host.rs
```diff
@@ -351,8 +352,7 @@ impl HostVmPrototype {
                     Ok(id)
                 },
             })?;
-            registered_functions.shrink_to_fit();
-            (vm_proto, registered_functions)
+            (vm_proto, registered_functions.into())
```
tomaka, line 355: Note that the conversion from `Vec` to `Arc<[]>` copies the elements, which is counter-balanced by the removal of the `shrink_to_fit`.

PR #1118, opened 2023-09-07
Publish v2.0.0 and new versions of the libraries

tomaka:


tomaka:
Actually, there's been a breaking change, so technically this should be v2.0.

Issue #519, opened 2023-05-04
Make the light-client more robust to panics by using `panic=unwind`


tomaka:
This isn't possible in Rust yet, but in principle we could use `panic=unwind` when compiling the wasm node. Then, the code could be tweaked in order to restart services if they die.

tomaka:
Unfortunately, using https://github.com/rust-lang/rust/pull/111322 seems incompatible with using `wasm32-wasi-threads` for #91 

Or maybe we could do threads manually instead of using `wasm32-wasi-threads`.

tomaka:
Ignoring the wasm-node, I think it's a good idea to do this in situations where the light client is embedded in a program where unwinding is enabled.

One blocker is the design of the "networking events receivers". It's not possible right now to create a new receiver in case the old one died in a service.

tomaka:
One issue with the latest comment is that it forces the `UnwindSafe` trait on `Platform`.
A bigger issue is (if I'm not mistaken) that `async` functions seem to generate futures that don't implement `UnwindSafe`.

tomaka:
The first thing to do would be to refactor the `RuntimeService` to have a clear separate between foreground and background (which is something I think should be done no matter what).

tomaka:
Here are the difficulties I've noted for each module:

- Network service: `lib.rs` injects bootnodes after creating the service, so if the network service crashes and restarts it won't know of any bootnode. Also the changes conflict with https://github.com/smol-dot/smoldot/issues/111.
- Runtime service: several functions need to return Results when otherwise they always succeed
- Transactions service: the channel that watches transactions has weird guarantees; wrap around another type to make things clearer
- Sync service: `serialize_chain_information` might be problematic
- JSON-RPC service: the whole design might need to be re-thought again in order to gracefully handle subscription tasks crashing

Another important point is that subscription IDs should be assigned by the frontend, in order to avoid race conditions.

tomaka:
After https://github.com/smol-dot/smoldot/pull/1376, I've tried to compile smoldot with the instructions given here: https://github.com/rust-lang/rust/pull/111322

This increases the binary size (expectedly) from 2.78 MiB to 3.0 MiB
I guess the trade-off could be worth it? Not totally sure.

tomaka:
Services to make panic-resilient:

- [ ] Network service
- [ ] Network service connection task
- [ ] Runtime service
- [ ] Transactions service
- [ ] Sync service
[...]

Issue #88, opened 2023-02-09
Use wasm64 for the wasmnode when available


tomaka:
https://github.com/paritytech/smoldot/issues/1789

tomaka:
I tried compiling for `wasm64-unknown-unknown` by curiosity.

The code of smoldot compiles just fine (after #1041), however when running with `node --experimental-wasm-memory64`, I get a weird error:

> Error while adding chain: CompileError: WebAssembly.compile(): Compiling function #19:"_ZN7smoldot6libp2p10read_write21ReadWrite$LT$TN..." failed: memory.copy[2] expected type i32, found local.get of type i64 @+14623

I don't know whether this is a bug in NodeJS or a miscompilation, but in both cases `wasm64` doesn't seem completely ready yet, and it's outside of our control.

PR #1129, opened 2023-09-08
Compile for `wasm32-unknown` instead of `wasm32-wasi`

tomaka:
Now that no-std compatibility has more or less been achieved (cc #133), we no longer need to compile to `wasi` and can instead switch to `wasm32-unknown-unknown`.

https://github.com/smol-dot/smoldot/issues/91 was preventing this change, but the issue is now closed.

Switching to `wasm32-unknown` paves the way for #88 or #519, and has the advantage that it's more stable than `wasi`. The bindings are also more clean. We no longer pin a specific version of Rust and no longer assume that `rustup` is in PATH and no longer install things on the user's machine, which are all great things.

The main disadvantage is that `println` will no longer work, which was sometimes useful for debugging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
       image: rust:1.72
     steps:
     - uses: actions/checkout@v4
+    - run: rustup target add wasm32-unknown-unknown
     - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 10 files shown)


tomaka:
Before this PR, `base64(gzip(wasm))` is 2.843 MiB
After this PR, `base64(gzip(wasm))` is 2.819 MiB

PR #1179, opened 2023-09-24
Move list of protocols to `peers.rs`

tomaka:
This PR is the continuation of https://github.com/smol-dot/smoldot/pull/462, and moves towards https://github.com/smol-dot/smoldot/issues/111

Instead of having the list of protocols in `collection.rs`, it is now found in `peers.rs`.
`collection.rs` now emits events about protocol negotiations that `peers.rs` must answer.

As always with the networking code, I'm not very happy about how it is, but I'm trying to define precise boundaries between the various modules before cleaning them up.


tomaka:
Node seems very stable.

Issue #1192, opened 2023-09-26
`TypeError` when calling `client.terminate()` after `Finalized block runtime ready...`


kratico:
`smoldot@2.0.2` throws `TypeError` when `client.terminate()` is invoked after a chain emits `Finalized block runtime ready...`

```
> node index.mjs

[smoldot] Smoldot v2.0.2
[smoldot] Chain initialization complete for westend2. Name: "Westend". Genesis hash: 0xe143…423e. Network identity: 12D3KooWDuSPNJDyUioToz1hCNvDHUQEyg9jMCYwsWrYPfwpTvJU. Chain specification starting at: 0x1deb…95c7 (#16847280)
[sync-service-westend2] GrandPa warp sync finished to #17629883 (0x9790…abb0)
[runtime-westend2] Finalized block runtime ready. Spec version: 9430. Size of `:code`: 1.2 MiB.
[smoldot] Shutting down chain westend2
TypeError: Cannot read properties of undefined (reading 'reset')
    at eventCallback (file:///home/projects/smoldot-panic-prujtf/node_modules/smoldot/dist/mjs/internals/client.js:180:28)
    at reset_connection (file:///home/projects/smoldot-panic-prujtf/node_modules/smoldot/dist/mjs/internals/local-instance.js:255:17)
    at wasm://wasm/0124b52e:wasm-function[4047]:0xe31f2
    at wasm://wasm/0124b52e:wasm-function[3524]:0xd91b8
    at wasm://wasm/0124b52e:wasm-function[2699]:0xbc110
    at wasm://wasm/0124b52e:wasm-function[1207]:0x3c96a
    at wasm://wasm/0124b52e:wasm-function[1206]:0x3c715
    at wasm://wasm/0124b52e:wasm-function[7062]:0x19567b
    at wasm://wasm/0124b52e:wasm-function[1230]:0x3e463
    at wasm://wasm/0124b52e:wasm-function[1229]:0x3e132
```

This may be unrelated, but the very same code that produces the above error, sometimes, get stuck and keeps emitting `GrandPa warp sync...`
```
> node index.mjs

[smoldot] Smoldot v2.0.2
[smoldot] Chain initialization complete for westend2. Name: "Westend". Genesis hash: 0xe143…423e. Network identity: 12D3KooWRLTACg76yPXX6kVi4fRiHazbx8Zo1SdwtEDmkBZjcNXF. Chain specification starting at: 0x1deb…95c7 (#16847280)
[...]

PR #1196, opened 2023-09-28
Catch `SIGINT` in the demo in order to call `remove` and `terminate`

tomaka:
cc https://github.com/smol-dot/smoldot/issues/1192


wasm-node/javascript/demo/demo.mjs
```diff
@@ -19,7 +19,7 @@
 
 import * as smoldot from '../dist/mjs/index-nodejs.js';
 import { WebSocketServer } from 'ws';
-import * as process from 'node:process';
+import process from 'node:process';
```
tomaka, line 22: Note that without this change I'm getting `process.on is not a function` errors. Have no idea why.

Issue #1090, opened 2023-08-30
Smoldot has panicked while executing task `network-service`.


Douglasacost:
Hello guys, recently I'm having some errors some seconds after a connection

As the error shows is panicked and should be reported, here the data

This is happening while getting connected to a parachain

<img width="883" alt="Screenshot 2023-08-30 at 8 14 27 AM" src="https://github.com/smol-dot/smoldot/assets/10488928/f7709fb1-a0bc-47bb-84c0-000820d318a1">

Node version: v18.17.0
Smoldot version: 1.0.17

tomaka:
Duplicate of https://github.com/smol-dot/smoldot/issues/381

(I've never taken the time to fix https://github.com/smol-dot/smoldot/issues/381 as it's not supposed to ever happen on a real chain)

tomaka:
I can't manage to reproduce. Do you have a reliable way of making it happen, by any chance?

tomaka:
As a side-note, this code is over-complicated, making it hard to figure out the problem. The plan is for #478 to simplify it.

skunert:
I also ran into this bug today on a zombienet test in CI. I think I have seen the error before but it seems to happen rarely.
Logs attached, but not very insightful sadly (Exported from loki, so starts from bottom).
[Explore-logs-2023-09-15 13_46_37.txt](https://github.com/smol-dot/smoldot/files/12619588/Explore-logs-2023-09-15.13_46_37.txt)

PR #1200, opened 2023-09-28
Rewrite `network::ChainNetwork` to not use the `peers` module

### LINKED ISSUE #111: Use only one network service for all chains

### LINKED ISSUE #381: Panic when inbound notifications substream fails to open
> Smoldot has panicked while executing task `network-service`. This is a bug in smoldot. Please open an issue at https://github.com/smol-dot/smoldot/issues with the following message:
panicked at 'called `Option::unwrap()` on a `None` value', /home/tomaka/Projets/smoldot/lib/src/network/service/notifications.rs:320:88

Can be reproduced for example by tweaking the protocol name to be intentionally invalid.

### LINKED ISSUE #478: Consider merging `service.rs` and `peers.rs` together
When thinking about how to update `peers.rs` in the context of https://github.com/smol-dot/smoldot/issues/111 and https://github.com/smol-dot/smoldot/issues/44, it seems to me that the logic `peers.rs` would be quite intertwined with the one in `service.rs`.
In other words, the way the logic of the networking is split between `peers.rs` and `service.rs` isn't really appropriate anymore.

It would make sense to me to remove `peers.rs` entirely, and move what it is doing within `service.rs`.

Of course, we can't just move all the fields of `peers.rs` within the service, as it would become way too complex, but we can add more state machines that help the service do what it has do to.

tomaka:
Close #478 
Close #111 
Close https://github.com/smol-dot/smoldot/issues/1090
Close #381 

Doesn't tackle #44, as it is a bit tricky to get right. This will be done in a further PR.

This PR also assumes that https://github.com/smol-dot/smoldot/pull/391 will be merged at some point. It is unclear at the moment whether #391 needs to be merged beforehand.

Remains to do, at the time of opening of this draft:

- Kademlia requests and Kademlia in general.
- The address book, which I'm not sure how to design.
- Transition the full-node and light-base network services to use the new service (should be very easy, as the API is mostly the same).
- Remove `peers.rs` and the old service (very easy).


tomaka:
I've more or less finished designing the API of the new service.

A few changes have been made compared to the current service:

- In and out slots are no longer handled directly by the service. The API user has to manually accept or deny each incoming gossip request, and "out slots" are now called "gossip desired".

- There's no concept of "pending connection" anymore. This previously designated connections that are still in their pre-libp2p initialization phase. There's actually no need to differentiate this phase, and we can just treat them as connections that are still handshaking. After all, the TCP handshake for example is also a handshake, just not handled directly by smoldot. If necessary, whether the remote is reachable can be determined by whether at least one byte has been received, but I don't really see why there would be a need to differentiate connections that are unreachable from connections that can't finish a handshake.

- `OutRequest` and `InRequest` are now `SubstreamId`.

- `KademliaOperationId` is now `OperationId`.

When it comes to peer addresses management, I'm going to keep the same algorithm(s) as right now and made changes to it after this PR has landed.

tomaka:
After reflection, I'm going to remove Kademlia and the address book from the service state machine in `lib`, and put them directly in the high-level network services.

tomaka:
The code in this PR is working surprisingly well.
The main thing that is broken is more or less Kademlia. Before this PR, we perform discovery by picking a random PeerId then querying the peer closest to this PeerId. After this PR, we pick a random PeerId but then ask any peer, not necessarily the closest. This leads to suboptimal results, but it's fine.

full-node/src/network_service/tasks.rs
```diff
@@ -42,45 +42,40 @@ use std::{
...
     address: String,
-    socket: impl AsyncReadWrite,
+    socket: impl Future<Output = Result<impl AsyncReadWrite, io::Error>>,
     connection_id: service::ConnectionId,
     mut connection_task: service::SingleStreamConnectionTask<Instant>,
     mut coordinator_to_connection: channel::Receiver<service::CoordinatorToConnection>,
     connection_to_coordinator: channel::Sender<super::ToBackground>,
 ) {
+    // Finishing ongoing connection process.
+    let socket = match socket.await.map_err(|_| ()) {
```
tomaka, line 56: This removes the timeout under the idea that the `ChainNetwork` will handle the timeout. This is true, except that this code doesn't actually process the messages from the `ChainNetwork`.

light-base/src/network_service/tasks.rs
```diff
@@ -16,178 +16,59 @@
...
 ) {
-    let mut socket = pin::pin!(socket);
+    let address_string = address.to_string();
+    let Ok(address_parse::AddressOrMultiStreamAddress::Address(address)) =
+        address_parse::multiaddr_to_address(&address)
+    else {
+        unreachable!()
+    };
+
+    let mut socket = pin::pin!(match platform.connect_stream(address).await {
```
tomaka, line 49: Same idea here

full-node/src/network_service.rs
```diff
@@ -842,185 +924,283 @@ async fn background_task(mut inner: Inner) {
...
+                                inner.network.gossip_remove_desired(
+                                    chain_id,
+                                    &peer_id,
+                                    service::GossipKind::ConsensusTransactions,
+                                );
+                                inner.peering_strategy.unassign_slot_and_ban(
+                                    &chain_id,
+                                    &peer_id,
+                                    Instant::now() + Duration::from_secs(10),
+                                );
```
tomaka, line 957: The gossip channel isn't actually closed here.

full-node/src/network_service.rs
```diff
@@ -842,185 +924,283 @@ async fn background_task(mut inner: Inner) {
...
+                                    peer_id, inner.chains[&chain_id].log_name, HashDisplay(&header_hash), decoded.is_best, error
                                 ));
 
-                                inner.unassign_slot_and_ban(chain_index, peer_id);
+                                inner.network.gossip_remove_desired(
+                                    chain_id,
+                                    &peer_id,
+                                    service::GossipKind::ConsensusTransactions,
+                                );
+                                inner.peering_strategy.unassign_slot_and_ban(
```
tomaka, line 953: Should log this

full-node/src/network_service.rs
```diff
@@ -842,185 +924,283 @@ async fn background_task(mut inner: Inner) {
...
-                                peer_id, chain_index
+                                "chain-disconnected; peer_id={}; chain={}",
+                                peer_id, inner.chains[&chain_id].log_name
                             ),
                         );
 
-                        inner.unassign_slot_and_ban(chain_index, peer_id.clone());
+                        // Note that peer doesn't necessarily have an out slot, as this event
+                        // might happen as a result of an inbound gossip connection.
+                        inner.peering_strategy.unassign_slot_and_ban(
```
tomaka, line 1000: Log

full-node/src/network_service.rs
```diff
@@ -842,185 +924,283 @@ async fn background_task(mut inner: Inner) {
...
+                            inner
+                                .peering_strategy
+                                .remove_chain_peer(&chain_id, &peer_id);
+                        } else {
+                            inner.peering_strategy.unassign_slot_and_ban(
+                                &chain_id,
+                                &peer_id,
+                                Instant::now() + Duration::from_secs(15),
+                            );
+                        }
```
tomaka, line 1047: Log this

full-node/src/network_service.rs
```diff
@@ -1058,43 +1238,44 @@ async fn background_task(mut inner: Inner) {
...
                     }
                     service::Event::ProtocolError { peer_id, error } => {
                         inner.log_callback.log(
                             LogLevel::Warn,
                             format!("protocol-error; peer_id={}; error={}", peer_id, error),
                         );
-                        for chain_index in 0..inner.network.num_chains() {
-                            inner.unassign_slot_and_ban(chain_index, peer_id.clone());
-                        }
+                        inner.peering_strategy.unassign_slots_and_ban(
```
tomaka, line 1275: Log this

full-node/src/network_service.rs
```diff
@@ -1129,41 +1310,38 @@ async fn background_task(mut inner: Inner) {
...
+                        >= 4
+                    {
                         break;
+                    }
+
+                    let peer_id = match inner.peering_strategy.assign_out_slot(chain_id, &Instant::now()) {
+                        basic_peering_strategy::AssignOutSlotOutcome::Assigned(peer_id) => {
+                            peer_id.clone()
+                        }
+                        basic_peering_strategy::AssignOutSlotOutcome::AllPeersBanned { .. }  // TODO: handle `AllPeersBanned` by waking up when a ban expires
```
tomaka, line 1328: Should be done before merging

full-node/src/network_service.rs
```diff
@@ -1176,40 +1354,108 @@ async fn background_task(mut inner: Inner) {
...
-                                info: start_connect,
-                            })
-                            .await;
+                let Some(multiaddr) = inner.peering_strategy.addr_to_connected(&peer_id) else {
+                    // There is no address for that peer in the address book.
+                    inner.network.gossip_remove_desired_all(
+                        &peer_id,
+                        service::GossipKind::ConsensusTransactions,
+                    );
+                    inner
```
tomaka, line 1370: Log this

full-node/src/network_service.rs
```diff
@@ -1176,40 +1354,108 @@ async fn background_task(mut inner: Inner) {
...
+                        .peering_strategy
+                        .unassign_slots_and_ban(&peer_id, Instant::now() + Duration::from_secs(10));
+                    continue;
+                };
+
+                let multiaddr = match multiaddr::Multiaddr::try_from(multiaddr.to_owned()) {
+                    Ok(a) => a,
+                    Err(multiaddr::FromVecError { addr }) => {
+                        // Address is in an invalid format.
```
tomaka, line 1379: Log this

tomaka:
Another thing to change: we don't actually need to track "in slots", as we can consider all the peers that we are connected to but don't have an "out slot" as automatically having an "in slot".
This can lead to a simplification of the peering strategy code.

tomaka:
I'm going to merge this PR 🎉 and open issues for the remaining things to fix, as these things to fix require some changes in other parts of the code and I'd rather not put changes to many different components in a single PR.

PR #1226, opened 2023-10-17
add method to expose storage_changes.trie_diffs

ermalkaleci:
We need this for chopsticks to get all the storage changes

### DIFF
--- a/lib/src/executor/runtime_host.rs
+++ b/lib/src/executor/runtime_host.rs
@@ -253,6 +253,12 @@ impl StorageChanges {
             .remove(&None)
             .unwrap_or(storage_diff::TrieDiff::empty())
     }
+
+    pub fn trie_diffs(
+        &self,
+    ) -> &hashbrown::HashMap<Option<Vec<u8>>, storage_diff::TrieDiff, fnv::FnvBuildHasher> {
+        &self.inner.trie_diffs
+    }
 }
 
 impl fmt::Debug for StorageChanges {


tomaka:
There are already functions that expose the list of changes: https://docs.rs/smoldot/latest/smoldot/executor/runtime_host/struct.StorageChanges.html#impl-StorageChanges

Is there something missing?

I'd rather not expose the `TrieDiff` type directly in the API.

ermalkaleci:
> There are already functions that expose the list of changes: https://docs.rs/smoldot/latest/smoldot/executor/runtime_host/struct.StorageChanges.html#impl-StorageChanges
> 
> Is there something missing?
> 
> I'd rather not expose the `TrieDiff` type directly in the API. (#780)

I need to read child storage diffs and I was using `trie_changes_iter_ordered` but somehow it gives me nothing when `child::kill` is performed. Having access to trie_diffs, I can iter and read all the storage diffs

tomaka:
`trie_changes_iter_ordered` wraps around `trie_diffs`, so if it doesn't yield anything that means that `trie_diffs` doesn't contain anything.

When you do `child::kill`, all that happens is that all the keys of the child trie are removed. If the child trie was empty, then this does nothing. Is it possible that the child trie was simply empty in the parent block?

ermalkaleci:
we're testing crowdloan refund which calls `child:kill` and it was refunding the same users because `trie_changes_iter_ordered` wasn't returning any storage diff for the killed child https://github.com/paritytech/polkadot/blob/52209dcfe546ff39cc031b92d64e787e7e8264d4/runtime/common/src/crowdloan/mod.rs#L708
not sure if there's some bug with `trie_changes_iter_ordered` but if I reading `trie_diffs` directly makes things much easier and seems to fix our issue

tomaka:
`StorageChanges` is indeed missing some functions to get the storage changes for child tries, however as mentioned I'd prefer to not expose `TrieDiff` in the API. I'd prefer to add more functions that allow introspecting the content of the `TrieDiff`s rather than return a reference to the `TrieDiff`s.

Also note that `StorageChanges` has a pretty good `fmt::Debug` implementation, so you can just print it out in order to be sure whether it contains an item or not.

ermalkaleci:
I'm happy with any solution. I can add a method that iter `trie_diffs` returning `(child, key, value)` or if you've better idea just let me know until then I'm using my fork so no rush.

PR #1250, opened 2023-10-27
Generalize `libp2p::with_buffers` for wasm environments

lexnv:
This PR makes the `libp2p::with_buffers` module generic over the `TNow`.

The `std::time::Instant` is not compatible with wasm environments. 
As such, the `Instant` field is generic to reuse the implementation for wasm environments.

Would love to get your feedback on this 🙏 

For more context, this subxt implementation currently has to copy and maintain the `with-buffers` module: https://github.com/paritytech/subxt/pull/1212.

cc @tomaka @jsdw


tomaka:
Making the `Instant` generic sounds reasonable to me, but I don't understand the point of the feature flag?
Especially given that you don't conditionally-compile anything depending on whether it's enabled.

lib/src/libp2p/with_buffers.rs
```diff
@@ -33,12 +33,12 @@ use core::{
...
 };
 use futures_util::{AsyncRead, AsyncWrite};
-use std::{io, time::Instant};
+use std::io;
 
 /// Holds an implementation of `AsyncRead` and `AsyncWrite`, alongside with a read buffer and a
 /// write buffer.
 #[pin_project::pin_project]
-pub struct WithBuffers<T> {
+pub struct WithBuffers<T, TNow> {
```
tomaka, line 41: Could you change `T` to `TSocket` everywhere, since there are two parameters now?

lexnv: Yep that makes sense thanks!

lexnv:
I've removed the `with-buffers` feature flag, it seems that `smoldot` compiles fine with `subxt` for wasm environments without it, thanks!

lib/src/libp2p/with_buffers.rs
```diff
@@ -314,15 +318,15 @@ where
     }
 }
 
-impl<T: fmt::Debug> fmt::Debug for WithBuffers<T> {
+impl<TSocket: fmt::Debug, TNow: Clone> fmt::Debug for WithBuffers<TSocket, TNow> {
```
tomaka, line 321:
```suggestion
impl<TSocket: fmt::Debug, TNow> fmt::Debug for WithBuffers<TSocket, TNow> {
```

tomaka (dismissed):

Issue #1244, opened 2023-10-27
Handle log level and target


xlc:
https://github.com/smol-dot/smoldot/blob/4faf2c043d31d3b814597dea8e95ba78e4a8b2dc/lib/src/executor/host.rs#L3438

I would love to have those to help with the log monitoring. For example, I want to trigger an alert for any logs with error level.

That also means this can't be a simple string, but a vector of some Log item

https://github.com/smol-dot/smoldot/blob/4faf2c043d31d3b814597dea8e95ba78e4a8b2dc/lib/src/executor/runtime_host.rs#L1230-L1231

tomaka:
> That also means this can't be a simple string, but a vector of some Log item

I would just generate an event for every log host function that's being called, like in `host.rs`

PR #1276, opened 2023-10-31
Make `runtime_host` stream the logs of the runtime

tomaka:
Close https://github.com/smol-dot/smoldot/issues/1244

cc @xlc 

As mentioned in https://github.com/smol-dot/smoldot/issues/1244, I went with the solution of adding a `LogEmit` variant to `RuntimeHostVm`. This allows introspecting the log message that the runtime wants to emit and react appropriately if desired.

I went with the opinionated decision of ignoring logs everywhere, because they're currently not printed anyway. Block body verification passes the logs through, so that the full node can eventually print them in case of block verification failure.


xlc:
Thanks. Will try to integrate this and let you know if I need anything else.

Issue #95, opened 2023-02-09
Treat clippy warnings as errors


tomaka:
https://github.com/paritytech/smoldot/issues/1623

PR #1290, opened 2023-11-02
Fix clippy warnings

Daanvdplas:
cc #95

The clippy warnings still left as for now:
- Checks for public functions that return a `Result` with an `Err` type of `()`. It suggests using a custom type that: implements `std::error::Error`
https://rust-lang.github.io/rust-clippy/master/index.html#/result_unit_err
- Checks for types used in structs, parameters and `let` declarations above a certain complexity threshold:
https://rust-lang.github.io/rust-clippy/master/index.html#/type_complexity
- Checks for functions with too many parameters:
https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments

And one which I'm not sure whether the clippy warning is correct ([see here](https://github.com/smol-dot/smoldot/blob/e7cb0b0c0d1389939939f2dfa4c35337da36dfc0/full-node/bin/cli.rs#L97)) - Checks for usage of Vec<Box<T>> where T: Sized anywhere in the code:
https://rust-lang.github.io/rust-clippy/master/index.html#/vec_box


lexnv (approved):

full-node/src/network_service.rs
```diff
@@ -859,17 +855,23 @@ async fn background_task(mut inner: Inner) {
...
-                                .peering_strategy
-                                .insert_or_set_connected_address(&peer_id, remote_addr.clone().into_vec(), 10) // TODO: constant
+                            if let basic_peering_strategy::InsertAddressResult::Inserted {
+                                address_removed: Some(addr_rm),
+                            } = inner.peering_strategy.insert_or_set_connected_address(
+                                &peer_id,
+                                remote_addr.clone().into_vec(),
+                                10,
+                            )
+                            // TODO: constant
```
tomaka, line 865:
```suggestion
                            if let basic_peering_strategy::InsertAddressResult::Inserted {
                                address_removed: Some(addr_rm),
                            } = inner.peering_strategy.insert_or_set_connected_address(
                                &peer_id,
                                remote_addr.clone().into_vec(),
                                10, // TODO: constant
                            )
```

Not sure if the suggestion passes rustfmt, but right now this TODO is detached from what it points to

full-node/src/network_service.rs
```diff
@@ -1176,16 +1178,22 @@ async fn background_task(mut inner: Inner) {
...
-                                    .peering_strategy
-                                    .insert_chain_peer(chain_id, peer_id.clone(), 100) // TODO: constant
+                                if let basic_peering_strategy::InsertChainPeerResult::Inserted {
+                                    peer_removed: Some(peer_removed),
+                                } = inner.peering_strategy.insert_chain_peer(
+                                    chain_id,
+                                    peer_id.clone(),
+                                    100,
+                                )
+                                // TODO: constant
```
tomaka, line 1188:
```suggestion
                                    100, // TODO: constant
                                )
```

tomaka (approved):

full-node/src/network_service.rs
tomaka: Modifying the TODO to be something like `// TODO: 10 is an arbitrary constant` also works if it's annoying

Daanvdplas: Somewhere in the process I probably moved it. They are now fixed and I run `cargo (+nightly) fmt`

tomaka (approved):
Thank you

PR #1312, opened 2023-11-09
feat: add crashCallback

kratico:
Add a `crashCallback` that will enable to detect a `smoldot` client crash more easily.

Currently, the `CrashError` can be detected by invoking the following methods
- `client.addChain`
- `chain.sendJsonRpc`
- `chain.nextJsonRpcResponse`
- `chain.remove`
- `client.terminate`

Detecting a crash on these methods is not very convenient.

I'm currently using a single `smoldot` instante in a browser extension that add many chains.
When there is a crash, some of these chains need to be recreated with a new `smoldot` client instance.
This API will facilitate this scenario.

cc @josepot

### DIFF
--- a/wasm-node/javascript/src/internals/client.ts
+++ b/wasm-node/javascript/src/internals/client.ts
@@ -281,6 +281,8 @@ export function start(options: ClientOptions, wasmModule: SmoldotBytecode | Prom
                     error: new CrashError(event.message),
                 }
 
+                options.crashCallback?.(state.instance.error)
+
                 state.connections.forEach((connec) => connec.reset());
                 state.connections.clear();
 
--- a/wasm-node/javascript/src/public-types.ts
+++ b/wasm-node/javascript/src/public-types.ts
@@ -321,6 +321,13 @@ export interface ClientOptions {
      * supported anyway.
      */
     forbidWebRtc?: boolean;
+
+    /**
+     * Callback that the client will invoke when the client crashes.
+     * 
+     * @param error the crash error
+     */
+    crashCallback?: (error: CrashError) => void
 }
 
 /**


tomaka:
I don't want "crashing" to be a normal thing. That would be like adding a `onTypeError` callback to a JavaScript library.

I would also strongly recommend you don't add complicated code to recreate a smoldot instance in case in crashes. You also don't add `try-catch` blocks when you call a third-party library, it's the same situation here.

josepot:
> Let's say you try to connect to Rococo and smoldot crashes. What even is the point of restarting smoldot? It's going to crash again in exactly the same way one second later.

You restart it again without Rococo and the rest of the chains work 🌈

tomaka:
> You restart it again without Rococo and the rest of the chains work 🌈

How would you know that it's Rococo causing the crash?

josepot:
> > You restart it again without Rococo and the rest of the chains work 🌈
> 
> How would you know that it's Rococo causing the crash?

Initially you initialize all the networks in parallel and you hope for the best. If shit hits the fan, then you try starting them sequentially: meaning that you only do `addChain` on the next one once you have received the "finalized" event of the current one.

If you get a crash before receiving a `finalized` block, then you know that there must be something wrong with that chain, so you exclude it from the equation and you start the "sequential" process again.

This is for the background process of the extension, which is trying to keep the chains synced so that dApps have a shorter waiting period. So, it is worth it to do those kinds of shenanigans in there.

That being said, I did tell @kratico that this should be a very low priority thing in our backlog, because well, as you said: smoldot shouldn't be panicking like that in the first place.

However, I do think that it could be quite handy to have an API on smoldot which allows its consumers to know that there have been a panic in the smoldot instance. I mean, even if you didn't want to have a complex logic like the one that I just described, even in the case that the consumer just wants to try again after a certain amount of time, etc.

I would understand your opposition to adding this API if it was going to add complexity on smoldot for not good reason, but this is 2 lines of code and it's not a breaking change.

tomaka:
> Initially you initialize all the networks in parallel and you hope for the best. If shit hits the fan, then you try starting them sequentially: meaning that you only do addChain on the next one once you have received the "finalized" event of the current one.

> If you get a crash before receiving a finalized block, then you know that there must be something wrong with that chain, so you exclude it from the equation and you start the "sequential" process again.

Do you realize that this is extremely wonky, as a crash can come from many different reasons, such as a peer sending a networking message, and that this mechanism is pretty complex, is very hard to test, and thus will probably be itself buggy?

> However, I do think that it could be quite handy to have an API on smoldot which allows its consumers to know that there have been a panic in the smoldot instance. I mean, even if you didn't want to have a complex logic like the one that I just described, even in the case that the consumer just wants to try again after a certain amount of time, etc.

> I would understand your opposition to adding this API if it was going to add complexity on smoldot for not good reason, but this is 2 lines of code and it's not a breaking change.

If you do care about recovering from issues in smoldot, you should also add `try-catch` blocks around every single call to a JavaScript function of smoldot, because there might be bugs in smoldot's JavaScript as well, and exceptions might be thrown.
Assuming that you put `try-catch` blocks around every single call to smoldot's JavaScript, then you will catch the `CrashError` and you can handle it the same way as for example a `TypeError`.

tomaka:
Closing this, as I'm against the idea, and I think that the fundamental way to solve the problem is #519

Issue #1308, opened 2023-11-08
Smoldot has panicked while executing task `sync-service-rococo_v2_2`


kratico:
Getting this error
```
Smoldot has panicked while executing task `sync-service-rococo_v2_2`. This is a bug in smoldot. Please open an issue at https://github.com/smol-dot/smoldot/issues with the following message:
panicked at 'assertion failed: curr.start_slot_number <= parent_slot_number', lib/src/verify/babe.rs:279:9
```

With the following script shared in [stackblitz](https://stackblitz.com/edit/smoldot-panic-i7pxqr?file=index.mjs)

```js
import { start } from "smoldot"
import chainSpec from "./rococo_v2_2.json" assert { type: "json" }

const client = start()

const chain = await client.addChain({ chainSpec: JSON.stringify(chainSpec) })

chain.sendJsonRpc(
  JSON.stringify({
    jsonrpc: "2.0",
    id: "id",
    method: "chainHead_unstable_follow",
    params: [true],
  }),
)

while (true) {
  try {
    console.log(await chain.nextJsonRpcResponse())
  } catch (error) {
    console.error(error)
    break
  }
}

chain.remove()

client.terminate()

``` 

I've checked this in [stackblitz](https://stackblitz.com/edit/smoldot-panic-i7pxqr?file=index.mjs) with `Node v18.18.0` and the following `smoldot` versions 
- `v2.0.7`,  `panicked at 'assertion failed: curr.start_slot_number <= parent_slot_number', lib/src/verify/babe.rs:280:9`
- `v2.0.6` and `v2.0.3`, stuck with `[sync-service-rococo_v2_2] GrandPa warp sync idle at block #7527149 (0x0x6cc8…d74a)`

Then, testing locally with OSX and `Node v18.18.0`, `smoldot` versions `v2.0.7`, `v2.0.6` and `v2.0.3` panic.

Note: I didn't have this issue on Monday with `smoldot@v2.0.3` so it might be related to a change in Rococo.

pepoviola:
Hi @kratico, I think is related to a current issue in `rococo` (the chain is currently stalled because of a wrong upgrade).

kratico:
Thx @pepoviola 

Follow up question, Should the `smoldot` client crash if a single chain is stalled?

kratico:
FYI, I'm using `smoldot` with many chains.
It is unfortunate that a single chain could make the `smoldot` client crash.

What would be the right way to recover from that?

[...]

PR #1317, opened 2023-11-10
Add more checks for the validity of the chain information

tomaka:
cc https://github.com/smol-dot/smoldot/issues/1308


tomaka:
Converted back to draft. Should also do Grandpa.

tomaka:
> Converted back to draft. Should also do Grandpa.

I've added a TODO instead, as it's a bit more complicated to do, and I kind of want to revisit this chain information system altogether at some point instead.

PR #1325, opened 2023-11-12
Add `ChainNetwork::remove_chain`

tomaka:
cc #111


tomaka:
It is actually a bit more complicated than this PR does. Non-block-announce substreams (in particular inbound ones) can still be opened without the API user's knowledge.

PR #1358, opened 2023-11-18
Ban peer after disconnect

tomaka:
We currently don't ban peers when we fail to connect to them, leading to smoldot trying over and over again to connect to the same address.

Work in progress, because I need the changes in https://github.com/smol-dot/smoldot/pull/1357


tomaka:
The logic of this is actually a bit weird, because there's a phase where peers are in `unconnected_desired()` even though no `Disconnected` event has been generated.

PR #1362, opened 2023-11-19
statemine proof_decode fails

ermalkaleci:
We use `decode_proof` on chopsticks to decode current validation data, modify storage value and create new proof for the new block. This method worked fine and still works with current statemine but 3 days ago it failed. This is a test with proof nodes from block 5,800,000 which fails and one with recent block which works fine. Any idea why it fails and if it can happen again?


tomaka:
Since a proof can contain multiple trie roots (main trie and child trie), the proof decoding code takes all the nodes that are decodable, then removes the ones that are children of another proof entry. The ones that remain are considered as trie roots.

Storage values are also proof entries, but usually they aren't decodable so it's not a problem.
However, here, entry `e80300000090010000900100000000000000000001d198d83e299dc6445b36551102a4b2eaa47abacd3b0657b5b1e498475cd9dd2b005039278c0400000000000000000000005039278c0400000000000000000000` happens to be a decodable storage value. It is decodable, but its children aren't, hence the error.

I did it this way in order to avoid having to hardcode in the proof decoding code the child tries prefix (`:child_trie:default:`). If that entry was the storage value of a key that starts with `:child_trie:default:`, then the error would be legitimate.

tomaka:
This is a situation that can now happen as a result of the trie V1 format. It wasn't a problem when I wrote the code, and then later I didn't think about this.

There are two different ways of fixing this:

- Hardcode the child trie proof in the proof decoding.
- Treat errors as non-fatal unless all the trie nodes in the proof are invalid.

I think I would go for the latter.

ermalkaleci:
> Since a proof can contain multiple trie roots (main trie and child tries), the proof decoding code takes all the nodes that are decodable, then removes the ones that are children of another proof entry. The ones that remain are considered as trie roots.
> 
> Storage values are also proof entries, but usually they aren't decodable and so they are discarded. However, here, entry `e80300000090010000900100000000000000000001d198d83e299dc6445b36551102a4b2eaa47abacd3b0657b5b1e498475cd9dd2b005039278c0400000000000000000000005039278c0400000000000000000000` happens to be a decodable storage value. It is decodable, but its children aren't, hence the error.
> 
> I did it this way in order to avoid having to hardcode in the proof decoding code the child tries prefix (`:child_trie:default:`). If that entry was the storage value of a key that starts with `:child_trie:default:`, then the error would be legitimate.

I did a little digging but didn't understand why there were 2 root hashes, now it makes sense

PR #1376, opened 2023-11-21
Make the `wasm-node` no-std

tomaka:
The `wasm-node` is now `no_std`.
This is ensured by the `#[panic_handler]`, which would fail to compile if a dependency wasn't no-std-friendly.

No change in behavior, as we just replaced code by what was done underneath. Note that the executor loses its multithreadedness property. The work-stealing queue of `async-executor` has been replaced with a simple FIFO queue, which is more appropriate in single-threaded contexts.


tomaka:
Change in the .wasm size:
Before this PR: 2.794 MiB
After this PR: 2.782 MiB

In other words, it's negligible.

lexnv (approved):

PR #1402, opened 2023-11-24
Make the runtime service panic-resilient

tomaka:
cc #519 

Makes the runtime service panic-resilient, following the same model as https://github.com/smol-dot/smoldot/pull/1401


tomaka:
Actually, I'm not a fan of the new `Crashed` error in the runtime, as I think it's too confusing. All the other errors are "pure" (based only on the input) except this one. This makes for example caching the runtime complicated.

PR #1445, opened 2023-12-01
Fix panic in optimistic sync verification queue

tomaka:
Only concerns the full node.

The panic comes from a confusion between "maximum block" and "one past maximum block".


tomaka:
The panic is in fact not fixed:

```
thread 'tasks-pool-6' panicked at /Users/tomaka/Projets/smoldot/lib/src/sync/optimistic/verification_queue.rs:134:43:
attempt to subtract with overflow
```

Issue #1446, opened 2023-12-01
SQLite database insertions very slow


tomaka:
This specific query is extremely slow: https://github.com/tomaka/smoldot/blob/65dd0a9b525946bf6413ad71c04c97b2ddbf829d/lib/src/database/full_sqlite.rs#L1397

The more blocks are in the database, the slower it becomes, which makes me think that there's an index missing.

tomaka:
```
QUERY PLAN
|--MATERIALIZE node_with_key
|  |--SETUP
|  |  |--MATERIALIZE insertions
|  |  |  |--SETUP
|  |  |  |  `--SCAN temp_pending_parent_copies
|  |  |  `--RECURSIVE STEP
|  |  |     |--SCAN insertions
|  |  |     |--SEARCH trie_node USING INDEX sqlite_autoindex_trie_node_1 (hash=?)
|  |  |     |--SEARCH trie_node_child USING INDEX trie_node_child_by_child_hash (child_hash=?) LEFT-JOIN
|  |  |     `--SEARCH trie_node_storage USING INDEX trie_node_storage_by_trie_root_ref (trie_root_ref=?) LEFT-JOIN
|  |  |--SEARCH blocks USING INDEX sqlite_autoindex_blocks_1 (hash=?)
|  |  |--SEARCH trie_node USING INDEX sqlite_autoindex_trie_node_1 (hash=?)
|  |  `--SCAN insertions
|  `--RECURSIVE STEP
|     |--SCAN node_with_key
|     |--SEARCH trie_node_child USING INDEX sqlite_autoindex_trie_node_child_1 (hash=? AND child_num=?)
|     `--SEARCH trie_node USING INDEX sqlite_autoindex_trie_node_1 (hash=?)
|--SCAN node_with_key
`--SEARCH trie_node_storage USING INDEX sqlite_autoindex_trie_node_storage_1 (node_hash=?)
```

Everything seems good.

tomaka:
Issue seems to be that `insertions` (and thus `node_with_key`) contains more and more elements the more blocks are in the database.

tomaka:
> The InsertTrieNode type needs to be modified in order to be able to know the full key of the trie node.

Actually no, as we can determine full key by looking at the other entries inserted with the block.

PR #1447, opened 2023-12-01
Fix slowness when inserting block in database

tomaka:
Close https://github.com/smol-dot/smoldot/issues/1446


tomaka:
Tests are failing because we're trying to insert a trie node that already exists.
Since this is unrelated to this change, I'll bypass the issue and add a TODO.

tomaka:
It still slows down progressively, not sure why, but at a much better rate than before.

PR #1476, opened 2023-12-17
Make database capable of not having the complete storage

tomaka:
cc https://github.com/smol-dot/smoldot/issues/131

Right now, for each block in the database, either the storage is entirely present or it is entirely absent.

In lights of https://github.com/smol-dot/smoldot/issues/131#issuecomment-1855801742, it is desirable to allow downloading the storage of a block over time.
In order to achieve, the database now supports inserting individual trie nodes, the idea being that we progressively download the missing trie nodes and insert them in the database over time.
When accessing the storage of a block, the database will now try to fulfill the request with what it has, and return an error if it doesn't have enough.

This PR actually overall simplifies the API of the database, as it removes some unclear corner cases related to inserting the storage of a block.


tomaka:
This PR also fixes a bug where the `branch_nodes` parameter of `block_storage_next_key` wasn't working properly.
For this reason, some JSON-RPC tests are failing.

Issue #1442, opened 2023-12-01
Peers should be banned when a request fails or when they misbehave


tomaka:
When a networking request fails, or when it succeeds but the result is invalid, the peer needs to be disconnected from and banned.

I thought I had an issue open about this, but I can't find it.

For some aspects this is easy, but for some, such as an invalid justification, this is harder due to the asynchronousness of communications between services.

tomaka:
I don't think that the network services should automatically ban peers due to failed requests. The network services always send requests only because the sync/consensus service called a function specifically to send a request. Banning a peer as a result feels like a hidden side effect.

Instead, the networking service should expose the concept of banning on its public API, letting the sync/consensus service ban peers.

When the sync/consensus service detect that a peer has misbehaved, it should be immediately removed from the list of sources, and the networking service then bans it. In the meanwhile, any message concerning this peer is ignored.

tomaka:
> When the sync/consensus service detect that a peer has misbehaved, it should be immediately removed from the list of sources, and the networking service then bans it. In the meanwhile, any message concerning this peer is ignored.

I don't think we actually need to do that. It would add a lot of complexity, and isn't strictly necessary.

Issue #1409, opened 2023-11-25
Remove optimistic syncing


tomaka:
The optimistic syncing is vulnerable to long range attacks. It's simply not a good idea to start from 0 and go up the blocks, instead we should always start from the head of the chain and go down.

When it comes to the use case of building an archive node, we can download all the block headers from the head to 0, and then only start building the archive.

Blocked on #131

PR #1482, opened 2023-12-17
Ban peers when a request fails

tomaka:
cc #1442 

Work in progress. While I wouldn't mind merging as is (except for the CHANGELOG), the unused function warning makes the compilation fail.


light-base/src/sync_service/standalone.rs
```diff
@@ -1223,12 +1274,24 @@ impl<TPlat: PlatformRef> Task<TPlat> {
                             HashDisplay(&verified_hash),
                             error
                         );
+
+                        // TODO: ban peers that have announced the block
```
tomaka, line 1278: Should be done before merging

tomaka: I've changed my mind about the fact that this is needed before merging.
We obviously need the help of the sync state machine here, but its guarantees in corner cases (such as knowing which sources know about a bad block) are not very well defined. The sync state machine needs a general clean-up, which is currently blocked on #1409.

I don't think that banning peers that advertise bad blocks is critical, and so it's fine to leave it as a todo.

Issue #93, opened 2023-02-09
Full node should load consensus data from chain storage instead of storing it in database


tomaka:
https://github.com/paritytech/smoldot/issues/1752

tomaka:
Blocked on https://github.com/smol-dot/smoldot/issues/131

PR #1483, opened 2023-12-17
Implement warp syncing for the full node

tomaka:
cc #131 

This big PR is a working change, and I'm opening it mostly to show progress.
I intend to extract individual changes from it and merge these individual changes separately.


tomaka:
The current state of this PR is that the full node successfully performs a warp sync, then verifies another block afterwards.
No further block is verified, and the node seems to freeze, likely due to a bug somewhere in the syncing code which I haven't investigated.

The code that verifies blocks is implemented in a hacky way, as it currently always downloads a call proof of executing the block, and panics if that proof is invalid.

Another issue is that when restarting the node the next time, the database doesn't contain the `:code` or `:heappages`, which leads to a crash. This could be solved by storing the code and heappages (and other storage items) that are downloaded during the warp syncing. This is maybe not a super robust solution, because, after https://github.com/smol-dot/smoldot/issues/93, if we ever modify the runtime calls that are being performed, the database might not contain them.

I'm not 100% sure how to continue. I was originally planning to create a "catch up state download service", however downloading the state requires knowing which peer knows which block, which in turn requires the help of the consensus service.
While it bloats a bit the consensus service, I feel like it is the responsibility of the consensus service to download the state.

tomaka:
> likely due to a bug somewhere in the syncing code which I haven't investigated.

This is actually a networking service service, as increasing the size of the buffer of events solves the problem.

tomaka:
Blocks verification fails with:

> Log { log_level: 0, target: "runtime", message: "panicked at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/collections/btree/navigate.rs:535:47:\ncalled `Option::unwrap()` on a `None` value" }
[2023-12-28T10:13:23.011Z] [warn] failed-block-verification; hash=0x0d76…7690; height=18793982; total_duration=797.057541ms; error=wasm `unreachable` instruction executed

I don't actually understand what's going on here.

tomaka:
Superseded by https://github.com/smol-dot/smoldot/pull/1538

PR #1485, opened 2023-12-18
Implement full mode in `all_forks.rs`

tomaka:
Change extracted from https://github.com/smol-dot/smoldot/pull/1483 then expanded further.

This implements "full mode" in `all_forks.rs`.
While "full mode" seems like an elaborate thing, all it actually does is making sure that we download the body of a block before proposing it for verification. The PR also adds a function to fetch that block body that has been downloaded. The rest is out of scope of that module.

This essentially finishes the API of `all_forks.rs`.

### DIFF
--- a/lib/src/sync/all_forks.rs
+++ b/lib/src/sync/all_forks.rs
@@ -42,14 +42,6 @@
 //! >              block. This is necessary in case it is this other block 5 that will end up
 //! >              being finalized.
 //!
-//! # Full vs non-full
-//!
-//! The [`Config::full`] option configures whether the state machine only holds headers of the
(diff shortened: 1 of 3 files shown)


lib/src/sync/all_forks.rs
```diff
@@ -1204,6 +1148,8 @@ impl<TBl, TRq, TSrc> FinishAncestrySearch<TBl, TRq, TSrc> {
             return Err((AncestrySearchResponseError::UnexpectedBlock, self.finish()));
         }
 
+        // TODO: /!\ must verify that scale_encoded_extrinsics matches the header's extrinsics root
```
tomaka, line 1151: Should implement that before merging

PR #1488, opened 2023-12-18
Update to wasmi v0.32 and add lazy validation

tomaka:
cc #864 

This PR shouldn't be merged as-is. The compilation mode should not always be `Lazy`. Instead, we should add a new `ExecHint`.

However I've done this change in order to measure how much faster it is.
To do so, I've measured three times how long it takes to compile the Westend runtime.

"Compile the Westend runtime" includes decoding the zstd-compressed runtime, parsing some custom Wasm sections, creating a `Module` (with wasmi), and creating an instance (with wasmi) but not executing it. So the times below are not just wasmi, but should still mainly be wasmi.

With wasmi v0.31:

264.161ms
195.032ms
210.432ms

With wasmi master branch (commit 86d097623592dc499852ea4bb01cace0097d70a4):

702.528ms
661.07ms
684.553ms

With wasmi master branch (commit 86d097623592dc499852ea4bb01cace0097d70a4) with compilation mode set to `Lazy` (this PR)

83.617ms
84.843ms
85.62ms

cc @Robbepop

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1636,6 +1636,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "multi-stash"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "685a9ac4b61f4e728e1d2c6a7844609c16527aeb5e6c865915c08e619c16410f"
+
+[[package]]
 name = "nix"
 version = "0.27.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2850,10 +2856,10 @@ dependencies = [
 
 [[package]]
 name = "wasmi"
-version = "0.31.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acfc1e384a36ca532d070a315925887247f3c7e23567e23e0ac9b1c5d6b8bf76"
+version = "0.32.0-beta.0"
+source = "git+https://github.com/paritytech/wasmi#86d097623592dc499852ea4bb01cace0097d70a4"
 dependencies = [
+ "multi-stash",
  "smallvec",
  "spin",
  "wasmi_arena",
@@ -2864,14 +2870,12 @@ dependencies = [
 [[package]]
 name = "wasmi_arena"
 version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "401c1f35e413fac1846d4843745589d9ec678977ab35a384db8ae7830525d468"
+source = "git+https://github.com/paritytech/wasmi#86d097623592dc499852ea4bb01cace0097d70a4"
 
 [[package]]
 name = "wasmi_core"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dcf1a7db34bff95b85c261002720c00c3a6168256dcb93041d3fa2054d19856a"
+version = "0.14.0"
+source = "git+https://github.com/paritytech/wasmi#86d097623592dc499852ea4bb01cace0097d70a4"
 dependencies = [
  "downcast-rs",
  "libm",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2867,8 +2867,9 @@ dependencies = [
 
 [[package]]
 name = "wasmi"
-version = "0.32.0-beta.1"
-source = "git+https://github.com/paritytech/wasmi#f96475ef1135d11858e0859cee97ea07f93c6d3d"
+version = "0.32.0-beta.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5aab06f5b4dadd225f47efe41308550ab01f0b6aa2585d7823c3f215e5777ebd"
 dependencies = [
  "multi-stash",
  "num-derive",
@@ -2883,12 +2884,14 @@ dependencies = [
 [[package]]
 name = "wasmi_arena"
 version = "0.4.0"
-source = "git+https://github.com/paritytech/wasmi#f96475ef1135d11858e0859cee97ea07f93c6d3d"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "401c1f35e413fac1846d4843745589d9ec678977ab35a384db8ae7830525d468"
 
 [[package]]
 name = "wasmi_core"
 version = "0.15.0"
-source = "git+https://github.com/paritytech/wasmi#f96475ef1135d11858e0859cee97ea07f93c6d3d"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3ac482df6761020b2b75c9aade41c105993c5b9f64156c349bb7ccad226a6ecd"
 dependencies = [
  "downcast-rs",
  "libm",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1686,6 +1686,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-derive"
+version = "0.4.1"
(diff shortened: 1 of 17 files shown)


Robbepop:
It is a bit weird that Wasmi `master` branch is roughly 3 times slower than Wasmi `v0.31.0` because in my own tests the slowdown was roughly 30%, not 200%.
Good though that lazy compilation actually improves the situation.

Robbepop:
@tomaka FYI: I just uploaded `v0.32.0-beta.1`.

Any explanation on your side for the severe slowdowns? Proper optimizations and optimized Wasm used?

tomaka:
What I measured is wasmi that has itself been compiled to .wasm and being executed by NodeJS.
Plus there is some overhead (the same no matter the version of wasmi) because the runtime is compressed and needs to be decompressed.
It's normal that it's slow and you shouldn't look at the number in absolute.

Robbepop:
> What I measured is wasmi that has itself been compiled to .wasm and being executed by NodeJS. Plus there is some overhead (the same no matter the version of wasmi) because the runtime is compressed and needs to be decompressed. It's normal that it's slow and you shouldn't look at the number in absolute.

I see that Wasmi `v0.31.0` takes roughly 200ms for the entire process and since you said that this mostly measures Wasmi I expected Wasmi `v0.32.0-beta.N` to take roughly 30% more, e.g. roughly 260ms. But in your tests it is WAY more, so i wonder why.
Also I expect lazy Wasm compilation for big Wasm modules to be roughly 10-25 times faster but in your case again it is just 8.5 times faster compared to the unrealistically slow measurements. So I am a bit confused about those numbers.

tomaka:
> Proper optimizations and optimized Wasm used?

Ok I wasn't actually optimizing the build (I'm a bit confused because the compilation options go through a couple of layers).
Running the same test later in an hour or two.

Robbepop:
> > Proper optimizations and optimized Wasm used?
> 
> Ok I wasn't actually optimizing the build (I'm a bit confused because the compilation options go through a couple of layers). Running the same test later in an hour or two.

Looking forward!

tomaka:
Okay here we go. I didn't actually wait an hour:

wasmi 0.31:

293.095ms
245.785ms
266.289ms

wasmi 0.32:

334.245ms
334.879ms
341.965ms

wasmi 0.32 with lazy compilation:

63.022ms
65.548ms
74.998ms

I've also uploaded the runtime (decompressed) in hexadecimal, if you want to try it yourself:

[runtime.wasm.txt](https://github.com/smol-dot/smoldot/files/13706457/runtime.wasm.txt)

Robbepop:
Thanks for the quick update!

The new numbers (especially eager) look way more realistic now.
The 5x speedup with lazy compilation is a bit low but I think this might be due to the other things that are still happening which you mentioned in the first post. Especially parsing custom sections is probably pretty slow.

tomaka:
> Especially parsing custom sections is probably pretty slow.

This is off-topic, but I don't think so. Each section in a wasm file starts with its length, so we can easily iterate between the (not so many) sections until we find the ones we want. And then the content of the sections (what I'm parsing) are somewhere around 100 to 200 bytes.

Robbepop:
> > Especially parsing custom sections is probably pretty slow.
> 
> This is off-topic, but I don't think so. Each section in a wasm file starts with its length, so we can easily iterate between the (not so many) sections until we find the ones we want. And then the content of the sections (what I'm parsing) are somewhere around 100 to 200 bytes.

Ah okay, 100-200 bytes indeed shouldn't make a huge difference.

tomaka:
@Robbepop Interestingly, one specific test seems to fail with `InvalidWasm("branching offset is out of bounds for wasmi bytecode")`. That Wasm was successfully accepted by wasmi 0.31.

I have many other tests in the repo that compile and run Wasm blobs, and these pass successfully. That specific failing test uses an old Polkadot runtime, so it's not impossible that the runtime was miscompiled or something.

Robbepop:
> @Robbepop Interestingly, one specific test seems to fail with `InvalidWasm("branching offset is out of bounds for wasmi bytecode")`. That Wasm was successfully accepted by wasmi 0.31.
> 
> I have many other tests in the repo that compile and run Wasm blobs, and these pass successfully. That specific failing test uses an old Polkadot runtime, so it's not impossible that the runtime was miscompiled or something.
> 
> You can find the failing Wasm here: https://github.com/smol-dot/smoldot/blob/main/lib/src/executor/vm/test-polkadot-runtime-v9160.wasm

Ah that's very interesting. I know what is happening there. Due to some optimizations the new Wasmi uses 16-bit encoded branching offsets whereas old Wasmi used 32-bit branching offsets. I would have never assumed that this was going to be problematic because it means that there are jumps over more than 32k Wasmi instructions. Note that a single Wasmi can make up to 4 Wasm instructions. This is a current limitation of the new Wasmi bytecode, however, if necessary I know how to deal with it.

tomaka:
> if necessary

Since a valid wasm bytecode fails to compile, it's kind of necessary, no?

Robbepop:
Every Wasm runtime has certain limitations and thus every wasm runtime may encounter valid Wasm files that fail to compile. So it is more of a question where to draw the line.

tomaka:
That's very surprising to me, especially after the discussion about lazy validation. It's like the WebAssembly committee/people care about some problematic situations but not others.

The Wasm runtime used in the failing test is one that was used on the Polkadot chain. It's not, say, a weird experiment. It's legitimate Rust code that was compiled with the Rust compiler. While the newer Polkadot runtime seems to work, the chances that a future runtime no longer compiles with wasmi seem relatively high to me.

Robbepop:
> That's very surprising to me, especially after the discussion about lazy validation. I understand that there are issues such as the maximum size of the stack, but I was convinced that validating WebAssembly would deterministically either pass or fail. It's like the WebAssembly committee/people care about problematic situations that are easy to solve but not the ones that are hard to solve.
> 
> The Wasm runtime used in the failing test is one that was used on the Polkadot chain. It's not, say, a weird experiment. It's legitimate Rust code that was compiled with the Rust compiler. While the newer Polkadot runtime seems to work, the chances that a future runtime no longer compiles with wasmi seem relatively high to me.

It is all about runtime limitations. For practical reasons those limitations should be chosen to allow for all practical use cases. Wasmtime for example has a limitation on function parameters, so even though a Wasm function with 2k parameters is valid it would fail to compile on Wasmtime. However, practical Wasm blobs dont make use of so many parameters anyways.

I was under the impression that 16-bit branch offsets should be enough for all practical use case but it seems I was wrong so I agree that we need to fix this.

Robbepop:
@tomaka https://github.com/paritytech/wasmi/issues/859

tomaka:
All tests are passing, apart from the ones that fail because git dependencies are forbidden.

Robbepop:
> All tests are passing, apart from the ones that fail because git dependencies are forbidden.

That's amazing! Thanks a lot for the testing and updates.

Robbepop:
@tomaka Just release `v0.32.0-beta.2`.

Robbepop:
@tomaka I see that you just started to use the new beta Wasmi. Is that a good idea? It isn't officially production ready, yet. If this is not a big deal for Smoldot then it is obviously great for the new Wasmi version to get battle tested.

tomaka:
All the runtimes that I've tried work, and given that changes in the runtime are usually pretty small, the chances that it breaks in the future are not very high.
It's not like I'm feeding untrusted wasm bytecode coming from the networking into wasmi for example, in which case it might be more "dangerous".
The worst case scenario if something is broken is that UIs don't work.

Robbepop:
@tomaka Okay great, thanks for the elaboration. I am happy to see that Smoldot is a perfect testing ground for the new Wasmi version. :)

PR #1527, opened 2023-12-28
Revert "Update to wasmi v0.32 and add lazy validation (#1488)"

tomaka:
Reverts https://github.com/smol-dot/smoldot/pull/1488

cc @robbepop

While working on https://github.com/smol-dot/smoldot/pull/1483, I noticed that some Wasm runtime calls fail with:

> panicked at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/collections/btree/navigate.rs:535:47:\ncalled Option::unwrap() on a None value

Reverting #1488 fixes this issue, so I'm going to revert for now.

The code that is panicking is very low-level, so I'm suspecting some kind of memory corruption or miscompilation or something like that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1635,12 +1635,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "multi-stash"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "685a9ac4b61f4e728e1d2c6a7844609c16527aeb5e6c865915c08e619c16410f"
-
-[[package]]
 name = "nix"
 version = "0.27.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


Robbepop:
Okay thanks for letting me know. Going to investigate as soon as possible.

tomaka:
To clarify, the panic in `btree/navigate.rs` comes from the Rust code that was compiled to Wasm, and executing this Wasm with wasmi 0.32 triggers it while executing it with wasmi 0.31 doesn't.

tomaka:
I could make a reproducible case, but it's a bit complicated because there are a lot of imported functions and a lot of calls to these imported functions.

Robbepop:
> I could make a reproducible case, but it's a bit complicated because there are a lot of imported functions and a lot of calls to these imported functions.

I would really appreciate that if you would do that but if not I can also try doing it. Very certainly going to be needed.

Robbepop:
@tomaka I just release Wasmi `v0.32.0-beta.3` with many bug fixes. I will continue the bug hunt but maybe this new version fixes the problems you encountered with one of the Polkadot runtimes?

https://github.com/paritytech/wasmi/releases/tag/v0.32.0-beta.3

tomaka:
It seems to work now!

PR #1548, opened 2024-01-03
Build the Docker image for `arm64` as well

tomaka:
Following the documentation: https://docs.docker.com/build/ci/github-actions/multi-platform/

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -47,6 +47,7 @@ jobs:
     - uses: docker/build-push-action@v5.1.0
       with:
         context: .
+        platforms: linux/amd64,linux/arm64
         file: ./full-node/Dockerfile
         load: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
         tags: ghcr.io/smol-dot/full-node:main


tomaka:
Well, this can't work because we explicitly specify a `--target` in the Dockerfile.

tomaka:
Unfortunately, it takes an entire hour to build the ARM64 image, due to the build running in an emulator. As of recently, GitHub has added ARM64 runners, but they seem to be only for enterprises.

tomaka:
cc https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

tomaka:
I believe that the code is now working, the only issue being that `docker manifest create` automatically pushes onto the registry and thus needs proper login.

The issue that the arm64 image takes an hour to build is unfortunately obviously problematic. Upgrading to a paid plan is not very costly by itself, but using the runner costs 0.16 € per minute, which is significant. An alternative could be to build the arm64 image only every now and then rather than at every PR.

tomaka:
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

PR #1551, opened 2024-01-03
Publish new versions to crates.io

tomaka:
cc @ermalkaleci


ermalkaleci:
Thank you. We actually use it as submodule, it's easier

Issue #1561, opened 2024-01-08
UnresolvedFunctionImport  for Moonbeam Runtimes


albertov19:
According to -> https://github.com/AcalaNetwork/chopsticks/issues/575

When using Moonbeam in the latest version of Chopsticks, and setting `allow-unresolved-imports` to `false`, you get:

```
npx @acala-network/chopsticks@latest -c moonbeam --allow-unresolved-imports false
Logs:

panicked at 'called `Result::unwrap()` on an `Err` value: VirtualMachine(UnresolvedFunctionImport { function: "ext_panic_handler_abort_on_panic_version_1", module_name: "env" })', src/task.rs:133:6

Stack:

Error
    at imports.wbg.__wbg_new_abda76e883ba8a5f (/home/purestake/.npm/_npx/81ad9c881cb83600/node_modules/@acala-network/chopsticks-executor/dist/cjs/chopsticks_executor.js:729:19)
    at wasm://wasm/009743be:wasm-function[728]:0x12aaea
    at wasm://wasm/009743be:wasm-function[886]:0x13420c
    at wasm://wasm/009743be:wasm-function[802]:0x12eaca
    at wasm://wasm/009743be:wasm-function[105]:0xa92ae
    at wasm://wasm/009743be:wasm-function[355]:0x10e846
    at wasm://wasm/009743be:wasm-function[937]:0x134a9e
    at __wbg_adapter_46 (/home/purestake/.npm/_npx/81ad9c881cb83600/node_modules/@acala-network/chopsticks-executor/dist/cjs/chopsticks_executor.js:407:10)
    at real (/home/purestake/.npm/_npx/81ad9c881cb83600/node_modules/@acala-network/chopsticks-executor/dist/cjs/chopsticks_executor.js:391:22)
    at node:internal/process/task_queues:141:7



node:internal/event_target:777
  process.nextTick(() => { throw err; });
                           ^
Error [RuntimeError]: unreachable
    at wasm://wasm/009743be:wasm-function[728]:0x12ac13
    at wasm://wasm/009743be:wasm-function[886]:0x13420c
    at wasm://wasm/009743be:wasm-function[802]:0x12eaca
    at wasm://wasm/009743be:wasm-function[105]:0xa92ae
    at wasm://wasm/009743be:wasm-function[355]:0x10e846
    at wasm://wasm/009743be:wasm-function[937]:0x134a9e
[...]

PR #1573, opened 2024-01-11
Add support for `ext_panic_handler_abort_on_panic_version_1`

tomaka:
Fix https://github.com/smol-dot/smoldot/issues/1561

https://github.com/paritytech/polkadot-sdk/blob/c8112e2c6fd9f1f5e7cb767b682fda746fc00520/substrate/primitives/io/src/lib.rs#L1570


lib/src/executor/host.rs
```diff
@@ -2142,7 +2142,7 @@ impl ReadyToRun {
                     return HostVm::Error {
                         error: Error::Utf8Error {
                             function: host_fn.name(),
-                            param_num: 2,
+                            param_num: 1,
```
tomaka, line 2145: Very small side fix. Usually I like to open separate PRs, but this is not worth it.

PR #1644, opened 2024-01-31
implement transaction-index host functions

ermalkaleci:
There're [2 pallets](https://github.com/search?q=repo%3Aparitytech%2Fpolkadot-sdk%20sp_io%3A%3Atransaction_index%3A%3Aindex&type=code) using [transaction-index](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/remark/src/lib.rs#L73) and some parachains seems to use it. It seems transaction-index is used only in [std](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/state-machine/src/overlayed_changes/mod.rs#L584) so I guess an empty implementation will be fine.

### DIFF
--- a/lib/src/executor/host.rs
+++ b/lib/src/executor/host.rs
@@ -2204,6 +2204,13 @@ impl ReadyToRun {
                     },
                 }
             }
+            HostFunction::ext_transaction_index_index_version_1
+            | HostFunction::ext_transaction_index_renew_version_1 => {
+                HostVm::ReadyToRun(ReadyToRun {
+                    inner: self.inner,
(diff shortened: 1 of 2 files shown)


ermalkaleci:
allow_unresolved_function: true will not throw the error on load but once its called it will throw the error

tomaka:
> once its called it will throw the error

They are offchain functions. The only proper way to implement them in smoldot is to return an error.

ermalkaleci:
ok I'm closing it

Issue #1651, opened 2024-02-01
AllForks should update WarpSync


tomaka:
When the all_forks finalizes blocks, the warp_sync state machine doesn't get updated, meaning that it will always perform a warp sync after some time.

PR #1656, opened 2024-02-02
Notify the warp_sync when the all_forks finalizes blocks

tomaka:
Fix https://github.com/smol-dot/smoldot/issues/1651

Before this PR, the warp sync state machine would simply run in parallel completely in isolation and perform a warp sync every 32 blocks.
This PR fixes this by notifying the warp sync state machine when the all_forks state machine finalizes.


tomaka:
Seems to work like a charm

PR #1647, opened 2024-02-01
fix `rpc_methods` return type

josepot:
My attempt at fixing #1646

PR #1657, opened 2024-02-02
Revert "fix `rpc_methods` return type (#1647)"

tomaka:
cc @josepot 
That PR broke PolkadotJS.

The point of `rpc_methods` was to be backwards compatible with the old JSON-RPC API.
If the JSON-RPC spec isn't backwards compatible, then it's the spec that is wrong.


josepot:
> cc @josepot That PR broke PolkadotJS.
> 
> The point of `rpc_methods` was to be backwards compatible with the old JSON-RPC API. If the JSON-RPC spec isn't backwards compatible, then it's the spec that is wrong.

Super sorry about this. I will open a PR to fix the spec, then.

PR #1667, opened 2024-02-06
Update back to wasmi v0.32

tomaka:
Revert https://github.com/smol-dot/smoldot/pull/1624


FlorianFranzen:
Not sure why building against an unstable version was included in 0.15.0, as it currently breaks builds again (currently pulls `v0.32.0-beta16`).

Issue #1415, opened 2023-11-26
Another JSON-RPC refactor


tomaka:
This would be the third refactor of the JSON-RPC service, but I'm slowly learning how to do it properly.

This is how it would look like:

- In `lib.rs`, add a "reverse proxy" that binds the socket (or equivalent) and the actual requests processor(s). The role of this reverse proxy is to maintain coherency from the socket point of view in case a requests processor crashes. It also has the responsibility of enforcing the number of subscriptions, to ensure equal balancing of requests when there is more than one JSON-RPC client, and to send unsubscribe requests if a client disconnects.

- In light-base and full node, add a "requests processor": a single task (no sub-tasks) that processes requests either immediately or by pushing them to a `FuturesUnordered`. From the point of view of this requests processor, there is only a single JSON-RPC client. This requests processor would be relatively simple, as it doesn't need to handle everything handled by the reverse proxy.

- In light-base and full node, connect the two with channels.

tomaka:
cc #519 and to some extend https://github.com/smol-dot/smoldot/issues/1382

tomaka:
Adding a single JSON-RPC reverse proxy (as in https://github.com/smol-dot/smoldot/pull/1436) seems really over-complicated.

However, we can split it in two:

- A "client sanitizer" that merges/discards notifications in case a JSON-RPC client is too slow, and also generates unsubscribe JSON-RPC requests when the client disconnects.
- A "server dispatcher" that redirects requests to servers, and properly handles when a server is removed.

PR #1685, opened 2024-02-15
Rewrite the light client JSON-RPC server

tomaka:
cc https://github.com/smol-dot/smoldot/issues/1415

This PR rewrites the light client JSON-RPC server in order to get rid of the `json_rpc::services` module of the library. This module was badly designed and is to be replaced with https://github.com/smol-dot/smoldot/pull/1436.

Once https://github.com/smol-dot/smoldot/pull/1436 is finished and merged, the code in this PR will be split into multiple different JSON-RPC servers (the idea right now being: one for the legacy API, one for the new API, one for transactions) that are merged together through the `ServersMultiplexer`.

The new code in this PR removes the security properties of the JSON-RPC server. They were never actually documented anywhere, and smoldot doesn't need to have a secure JSON-RPC server at the moment. These security properties will be restored after https://github.com/smol-dot/smoldot/pull/1436.

Work time: 45h


tomaka:
Note that I've been trying to keep the existing behavior as much as possible.

However, I remember instinctively changing some return values w.r.t. the new JSON-RPC API to better match the spec, but maybe not all of them. Unfortunately, I don't remember exactly what I've changed, so I don't know what to put in the CHANGELOG.
After this PR, I will do a review of the return codes so that they match the spec.

PR #1724, opened 2024-03-15
Add support for `transaction_v1_broadcast`

tomaka:
Adds support for the new JSON-RPC API functions.

Work time: 1h

### DIFF
--- a/light-base/src/json_rpc_service/background.rs
+++ b/light-base/src/json_rpc_service/background.rs
@@ -123,8 +123,13 @@ struct Background<TPlat: PlatformRef> {
     /// List of all active `state_subscribeRuntimeVersion` subscriptions, indexed by the
     /// subscription ID.
     runtime_version_subscriptions: hashbrown::HashSet<String, fnv::FnvBuildHasher>,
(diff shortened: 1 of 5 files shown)


light-base/src/json_rpc_service/background.rs
```diff
@@ -2553,11 +2585,44 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                                .await;
+                        } else {
+                            let _ = me
+                                .responses_tx
+                                .send(parse::build_error_response(
+                                    request_id_json,
+                                    json_rpc::parse::ErrorResponse::ApplicationDefined(
+                                        -32800,
+                                        "Invalid operationId",
+                                    ),
```
tomaka, line 2612: Note sure what the JSON-RPC error should be if `transaction_v1_stop` is called with an invalid id.

lexnv: I have returned `-32602` (invalid param error) in Substrate. We could probably clarify this in the spec

lexnv (approved):

PR #1750, opened 2024-04-05
Run `npm update`

tomaka:
This updates the lock file version.


tomaka:
I guess we need to solve the problem in https://github.com/smol-dot/smoldot/pull/1408
Why does this always have to be so complicated when it comes to JavaScript

Issue #887, opened 2023-07-11
Extract test fixtures from Rust files


tomaka:
In the context of https://github.com/smol-dot/smoldot/pull/882, we shouldn't put test fixtures in Rust files anymore. Instead, they should be converted to JSON or similar.

This concerns for example `chain_spec.rs`, but probably others as well.

tomaka:
The only other big one is in `trie/prefix_proof/tests.rs`.

PR #1773, opened 2024-04-18
Extract test fixture from prefix_proof/tests

tomaka:
Close https://github.com/smol-dot/smoldot/issues/887

Extracts test data, so that it doesn't get uploaded to crates.io.

Work time: 20mn


tomaka:
The diff is misleading, as all the values are now on a single line.

PR #1846, opened 2024-05-29
Put the LinkerBuilder in BaseComponents

### LINKED ISSUE #1779: Use new `wasmi::LinkerBuilder` to improve startup time (?)
Wasmi `v0.32.0-beta.9` introduced the new `LinkerBuilder` API which can be used to quickly spin up new `Linker` instances whenever a new `Engine` is created. The `LinkerBuilder` can be used to pre-define a set of host functions and create `Linker` instances with this same set of host functions a lot more quickly afterwards.

However, setting up a `LinkerBuilder` initially is as costly as setting up a `Linker` without `LinkerBuilder`.

I implemented this for the `contracts-pallet` which has to create a new `Engine` per smart contract call (unfortunately).
Spinning up a new `Linker` using `LinkerBuilder` improves performance of `Linker` setup with 52 host functions by a factor of 6.

I was wondering if this optimization could also be applied to `smoldot` to improve its startup time further.

From what I see this is where `smoldot` uses Wasmi's `Linker`:
https://github.com/smol-dot/smoldot/blob/60d66124b10ccb4927161630fafc7c3d0a6641ab/lib/src/executor/vm/interpreter.rs#L120

It creates host functions on demand, so it might either need some refactoring to make use of `LinkerBuilder` or won't profit from it a lot.

tomaka:
Close #1779

Work time: 15mn

### DIFF
--- a/lib/src/executor/vm/interpreter.rs
+++ b/lib/src/executor/vm/interpreter.rs
@@ -45,9 +45,8 @@ pub struct InterpreterPrototype {
 struct BaseComponents {
     module: Arc<wasmi::Module>,
 
-    /// For each import of the module, either `None` if not a function, or `Some` containing the
-    /// `usize` of that function.
-    resolved_imports: Vec<Option<usize>>,
+    /// All function imports of a module are stored here.
+    linker_builder: wasmi::LinkerBuilder<wasmi::state::Ready, ()>,
 }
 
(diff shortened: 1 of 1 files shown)


Robbepop:
@tomaka happy to see that smoldot is now using the new `LinkerBuilder`. Did you by any chance measure the performance impact of this change?

PR #1856, opened 2024-05-30
Replace cloudflare-ipfs.com with ipfs.io

tomaka:
CloudFlare's IPFS portal is very slow, and they have announced shutting down in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Where "crash" includes: Rust panics, JavaScript exceptions (except for the ones 
 
 In order to run the wasm light node, you must have installed [rustup](https://rustup.rs/).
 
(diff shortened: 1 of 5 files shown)


hitchhooker:
we got now ibp.network PoPs in 12 locations bringing global latencies to max 100ms in the geodns(anycast in process). would probably make a lot of sense to utilize this to deploy own ipns/ipfsgateway service. do you got any recommendation for client implementation?

tomaka:
I don't exactly understand what you're asking.

This PR just modifies the smoldot demo that I use for local testing. Smoldot itself doesn't contain any URL to any IPFS gateway or anything like that.

PR #1937, opened 2024-08-12
feat: JSON-RPC responses async iterable iterator

tien:
Add support for the [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) statement for handling responses.

Currently, to listen to all responses, you would use a `while` loop like this:

```ts
try {
  while (true) {
    const response = await chain.nextJsonRpcResponse();
    console.info(response);
  }
} catch (error) {
  if (error instanceof AlreadyDestroyedError) {
    console.info("No more responses");
  } else {
    console.error(error);
  }
}
```

This PR introduces a more concise syntax:

```ts
try {
  for await (const response of chain.jsonRpcResponses) {
    console.log(response);
  }
  console.info("No more responses");
} catch (error) {
  console.error(error);
}
```

In addition to the syntactic improvement and avoiding the `while` loop, this approach allows for closing the stream without requiring an error to be thrown.

### DIFF
--- a/wasm-node/CHANGELOG.md
+++ b/wasm-node/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add `jsonRpcResponses` async iterable iterator to `Chain` connection
+
 ### Fixed
 
 - Fix potential panic in parachain syncing code. ([#1912](https://github.com/smol-dot/smoldot/pull/1912))
--- a/wasm-node/CHANGELOG.md
+++ b/wasm-node/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Add `jsonRpcResponses` async iterable iterator to `Chain` connection
+- Add `jsonRpcResponses` async iterable iterator to `Chain`, as a more convenient alternative to the `nextJsonRpcResponse` function. ([#1937](https://github.com/smol-dot/smoldot/pull/1937))
 
 ### Fixed
 
--- a/wasm-node/javascript/src/internals/client.ts
+++ b/wasm-node/javascript/src/internals/client.ts
@@ -552,27 +552,41 @@ export function start(options: ClientOptions, wasmModule: SmoldotBytecode | Prom
                         default: throw new Error("Internal error: unknown json_rpc_send error code: " + retVal)
                     }
                 },
-                nextJsonRpcResponse: async () => {
-                    while (true) {
-                        if (!state.chains.has(chainId))
-                            throw new AlreadyDestroyedError();
-                        if (options.disableJsonRpc)
-                            return Promise.reject(new JsonRpcDisabledError());
-                        if (state.instance.status === "destroyed")
-                            throw state.instance.error;
-                        if (state.instance.status !== "ready")
-                            throw new Error(); // Internal error. Never supposed to happen.
-
-                        // Try to pop a message from the queue.
-                        const message = state.instance.instance.peekJsonRpcResponse(chainId);
-                        if (message)
-                            return message;
-
-                        // If no message is available, wait for one to be.
-                        await new Promise<void>((resolve) => {
-                            state.chains.get(chainId)!.jsonRpcResponsesPromises.push(resolve)
-                        });
+                jsonRpcResponses: {
+                    next: async () => {
+                        while (true) {
+                            if (!state.chains.has(chainId))
+                                return { done: true, value: undefined };
+                            if (options.disableJsonRpc)
+                                throw new JsonRpcDisabledError();
+                            if (state.instance.status === "destroyed")
+                                throw state.instance.error;
+                            if (state.instance.status !== "ready")
+                                throw new Error(); // Internal error. Never supposed to happen.
+    
+                            // Try to pop a message from the queue.
+                            const message = state.instance.instance.peekJsonRpcResponse(chainId);
+                            if (message)
+                                return { done: false, value: message };
+    
+                            // If no message is available, wait for one to be.
+                            await new Promise<void>((resolve) => {
+                                state.chains.get(chainId)!.jsonRpcResponsesPromises.push(resolve)
+                            });
+                        }
+                    },
+                    [Symbol.asyncIterator]() {
+                        return this
                     }
+                },
+                nextJsonRpcResponse: async () => {
+                    const result = await newChain.jsonRpcResponses.next();
+
+                    if (result.done) {
+                        throw new AlreadyDestroyedError();
+                    }
+
+                    return result.value;
                 },
                 remove: () => {
                     if (state.instance.status === "destroyed")
--- a/wasm-node/javascript/src/public-types.ts
+++ b/wasm-node/javascript/src/public-types.ts
@@ -178,6 +178,20 @@ export interface Chain {
     nextJsonRpcResponse(): Promise<string>;
 
     /**
+     * JSON-RPC responses or notifications async iterable.
+     *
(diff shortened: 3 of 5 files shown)


tomaka:
Thanks for the change, but this is unfortunately an (unnecessary) API breaking change so I wouldn't accept it.

wasm-node/javascript/src/public-types.ts
```diff
@@ -177,6 +177,14 @@ export interface Chain {
      */
     nextJsonRpcResponse(): Promise<string>;
 
+    /**
+     * JSON-RPC responses or notifications to be generated.
+     *
+     * @throws {@link JsonRpcDisabledError} If the JSON-RPC system was disabled in the options of the chain.
+     * @throws {@link CrashError} If the background client has crashed.
+     */
```
tomaka, line 185: This needs to mention that it's an async iterable, repeat what `nextJsonRpcResponse` mentions, and explain that iteration stops if the chain is removed.

wasm-node/javascript/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "target": "ES6",
```
tomaka, line 3: This is already in the two other `tsconfig` JSON files, there shouldn't be any need to repeat it here.

```suggestion
```

wasm-node/javascript/src/internals/client.ts
```diff
@@ -552,27 +552,45 @@ export function start(options: ClientOptions, wasmModule: SmoldotBytecode | Prom
...
 
-                        // If no message is available, wait for one to be.
-                        await new Promise<void>((resolve) => {
-                            state.chains.get(chainId)!.jsonRpcResponsesPromises.push(resolve)
-                        });
+                    if (result.done) {
+                        throw new AlreadyDestroyedError();
                     }
+
+                    return result.value;
```
tomaka, line 593: I would actually much prefer if this function was left as it is, and `jsonRpcResponses` was implemented on top of it. This function is much more simple than `jsonRpcResponses` from an implementation point of view.

wasm-node/javascript/src/internals/client.ts
```diff
@@ -552,27 +552,45 @@ export function start(options: ClientOptions, wasmModule: SmoldotBytecode | Prom
...
+                                return { done: false, value: message };
+    
+                            // If no message is available, wait for one to be.
+                            await new Promise<void>((resolve) => {
+                                state.chains.get(chainId)!.jsonRpcResponsesPromises.push(resolve)
+                            });
+                        }
+                    },
+                    return: async () => {
+                        newChain.remove();
```
tomaka, line 579: Why remove the chain here?

wasm-node/javascript/src/public-types.ts
```diff
@@ -177,6 +177,14 @@ export interface Chain {
...
+    jsonRpcResponses: AsyncIterableIterator<string>
```
tomaka, line 186: What about?

```suggestion
    readonly jsonRpcResponses: AsyncIterableIterator<string>
```

And turn `jsonRcpResponses` into a getter?

tomaka (changes requested):
👍

wasm-node/javascript/src/public-types.ts
tien: Yes, this make sense, I've explicitly added a comment marking this as an async iterable.

I have also copied the comment over from `nextJsonRpcResponse`, except for the part explaining simultaneous calls, because I believe this is an implied behaviour of `AsyncIterableIterator`. 

i.e.

```ts
// It's implied that 2 iterations happen, therefor each will yield different response
Promise.all([
  chain.jsonRpcResponses.next(),
  chain.jsonRpcResponses.next(),
]);
```

wasm-node/javascript/tsconfig.json
tien: Oops, an oversight on my part, thanks for pointing this out, have removed accordingly 🙏

wasm-node/javascript/src/internals/client.ts
tien: I believe this behaviour make sense according to the documentation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#returnvalue). But I have removed it anyway, let me know what is your final judgement on this, thank you.

wasm-node/javascript/src/public-types.ts
tien: I have made this a `readonly` property. I think a property make more sense compared to a getter, because there's in fact only 1 iterable here:
- Each iteration yield the next value
- If multiple iterables are created and are iterated on, they will actually overlap (similar to call `nextJsonResponse` simultaniously)

tomaka: What I meant by getter is to do `get jsonRpcResponses() { return { next: ..., ... } }`

tien: Right, here I believe just a TypeScript `readonly` modifier is sufficient, like technically speaking other properties/methods like `sendJsonRpc`, `nextJsonRpcResponse`, etc are actually mutable as well. If you want to enforce immutability at runtime on the JS side, I think calling `Object.freeze` on the entire `chain` object is more effective.

wasm-node/javascript/src/public-types.ts
```diff
@@ -177,6 +177,20 @@ export interface Chain {
      */
     nextJsonRpcResponse(): Promise<string>;
 
+    /**
+     * JSON-RPC responses or notifications async iterable.
+     *
+     * Each chain contains a buffer of the responses waiting to be sent out. Iterating over this
+     * pulls one element from the buffer. If the iteration happen at a slower rate than
+     * responses are generated, then the buffer will eventually become full, at which point further iteration on
+     * {@link Chain.jsonRpcResponses} will throw an exception. The size of this buffer can be configured
```
tomaka, line 186: It's `sendJsonRpc` that would throw an exception.

tomaka (changes requested):
Looks good overall, could also add an entry to wasm-node/javascript/CHANGELOG.md? 🙏

wasm-node/javascript/src/public-types.ts
tien: Oops, thanks for the catch!

tien:
> Looks good overall, could you also add an entry to wasm-node/javascript/CHANGELOG.md? 🙏

I've added an entry to the changelog, the file you reference doesn't seem to exist though, so I have created one accordingly 🙏

tomaka:
> I've added an entry to the changelog, the file you reference doesn't seem to exist though, so I have created one accordingly 🙏

It was in fact `wasm-node/CHANGELOG.md`

tien:
> > I've added an entry to the changelog, the file you reference doesn't seem to exist though, so I have created one accordingly 🙏
> 
> It was in fact `wasm-node/CHANGELOG.md`

Okay, thanks for clarifying, have added the entry to the correct file 💪

wasm-node/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add `jsonRpcResponses` async iterable iterator to `Chain` connection
```
tomaka, line 7:
```suggestion
- Add `jsonRpcResponses` async iterable iterator to `Chain`, as a more convenient alternative to the `nextJsonRpcResponse` function. ([#1937](https://github.com/smol-dot/smoldot/pull/1937))
```

tomaka (approved):

PR #2037, opened 2024-11-18
Fix #2035

### LINKED ISSUE #2035: `chainHead_v1_call` always yields `operationInaccessible` events for `DryRunApi_dry_run_call`
The `DryRunApi_dry_run_call` runtime-API always yields an `operationInaccessible` event, making it unusable with smoldot.

This runtime-api is very useful for calculating the XCM delivery fees and for many other use-cases.

This issue is preventing the [following PR](https://github.com/kheopswap/kheopswap/pull/40) from being merged.

I have created the [following repo](https://github.com/josepot/dry-run-smoldot-issue) to reproduce this issue, using [this code](https://github.com/josepot/dry-run-smoldot-issue/blob/main/index.ts) (the exact same code works with a WebSocket provider).

The logs that I captured after running that code:
- [Wire logs](https://github.com/josepot/dry-run-smoldot-issue/blob/51ebb523214cbdb84c74d68f51f4dc782fcdc5c3/wire-logs.txt)
- [Smoldot logs](https://github.com/josepot/dry-run-smoldot-issue/blob/51ebb523214cbdb84c74d68f51f4dc782fcdc5c3/smoldot-logs.txt)

I'm more than willing to test this problem with a custom build of smoldot and/or to provide any other assistance to get to the bottom of this issue.

tomaka:
Fix #2035

This took me some time, as I had to get a proper understanding of the code first, but I'm 99.9% confident that what my PR changes is just a stupid mistake as I used the wrong variable in this edge case.

Work time: 1h


lib/src/trie/proof_decode.rs
```diff
@@ -1218,11 +1220,11 @@ impl<T: AsRef<[u8]>> DecodedTrieProof<T> {
                             // If the child isn't present in the proof, then the proof is
                             // incomplete. While in some situations we could prove that the child
                             // is necessarily the next key, there is no way to know its full key.
-                            if children_present_in_proof_bitmap & (1 << key_before_nibble) == 0 {
+                            if children_present_in_proof_bitmap & (1 << child_num) == 0 {
                                 return Err(IncompleteProofError());
                             }
 
-                            for c in 0..key_before_nibble {
+                            for c in 0..child_num {
```
tomaka, line 1227: Note that this could be `key_before_nibble` as well, and in fact it would be slightly faster to keep `key_before_nibble` (we know that the range between `key_before_nibble` and `child_num` is always empty), but using `child_num` makes this code way less confusing.

PR #2046, opened 2024-11-23
fix race-condition with worker API

### LINKED ISSUE #2045: `RangeError [Error]` when using `startFromWorker` API
Sometimes, when using the `startFromWorker` API I get the following error:

image

In most cases this issue doesn't happen in a deterministic manner, but somehow I managed to create [a reproduction ](https://github.com/josepot/smoldot-worker-race-condition)where the issue does happen deterministically.

josepot:
Fixes #2045

### DIFF
--- a/wasm-node/javascript/src/internals/local-instance.ts
+++ b/wasm-node/javascript/src/internals/local-instance.ts
@@ -511,18 +511,18 @@ export async function startLocalInstance(config: Config, wasmModule: WebAssembly
 
             const mem = new Uint8Array(state.instance.exports.memory.buffer);
             const responseInfo = state.instance.exports.json_rpc_responses_peek(chainId) >>> 0;
-            const ptr = buffer.readUInt32LE(mem, responseInfo) >>> 0;
-            const len = buffer.readUInt32LE(mem, responseInfo + 4) >>> 0;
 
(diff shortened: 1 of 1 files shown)


wasm-node/javascript/src/internals/local-instance.ts
```diff
@@ -511,18 +511,20 @@ export async function startLocalInstance(config: Config, wasmModule: WebAssembly
 
             const mem = new Uint8Array(state.instance.exports.memory.buffer);
             const responseInfo = state.instance.exports.json_rpc_responses_peek(chainId) >>> 0;
-            const ptr = buffer.readUInt32LE(mem, responseInfo) >>> 0;
-            const len = buffer.readUInt32LE(mem, responseInfo + 4) >>> 0;
 
+            if (mem.byteLength < responseInfo + 8)
+                return null
+
+            const len = buffer.readUInt32LE(mem, responseInfo + 4) >>> 0;
```
josepot, line 518: In some instances, the buffer is initialized but it still doesn't have the data for the length in it. The new condition exists if the `len` can't be read, instead of throwing a runtime error.

josepot:
Closing it in favour of #2047

PR #2047, opened 2024-11-24
Fix using invalidated Wasm memory

### LINKED ISSUE #2045: `RangeError [Error]` when using `startFromWorker` API

tomaka:
Probably fixes #2045, but I haven't tested.

Work time: 10mn

### DIFF
--- a/wasm-node/CHANGELOG.md
+++ b/wasm-node/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


josepot:
> Probably fixes #2045, but I haven't tested.

It does fix it, thanks!

Issue #2128, opened 2025-05-20
Transactions broadcast via smoldot on Kusama are not being included


josepot:
We’re observing a strange issue when broadcasting transactions using smoldot on Kusama:

* Transactions are validated locally and broadcast successfully.
* Smoldot logs indicate the transaction is valid and announced.
* However, many of these transactions:

  * Do not get included in any block.
  * Never show up in public RPCs' transaction pools.
  * Eventually become invalid due to `AncientBirthBlock` (i.e., mortality expires after \~10 mins).
  * Or, in some cases, get included *very*, very late (despite low network activity).

Notably, when broadcasting the *same* transaction through a centralized RPC, it gets included almost immediately.

We suspect this might not be a smoldot issue, but there seems to be some issue with the way that the blocks are being announced on Kusama... 🤷 

* [Logs: transaction became invalid after a long time](https://github.com/user-attachments/files/20357141/not-included.txt)
* [Logs: transaction got included very late](https://github.com/user-attachments/files/20357147/included-very-late.txt)

Could you please take a look? 🙏

As of right now it's almost impossible to send transactions on Kusama using PAPI (with smoldot)

tomaka:
If you search the logs for "announced-to-network" (or "transaction-announced"), you can see that every 5 seconds smoldot sends the transaction to a few peers. The list of peers slowly rotates randomly, so that even if some of them are malfunctioning, it should still work at some point.
I don't really see how I could make this behavior better.

Maybe there's been some protocol change in Kusama which I'm not aware of (and the reason why it still works after a while is that some peers are on an old version)? Some quick search from me found [this PR](https://github.com/paritytech/polkadot-sdk/pull/8128) from April, where the "current behavior" is still the one I know of.
The next step to me would be to look at the logs of Substrate, or open an issue in Polkadot-SDK.

tomaka:
I'm going to suspect https://github.com/paritytech/polkadot-sdk/pull/7866

cc @lexnv 
Could you tell us if there was some kind of intentional protocol change in the transactions? 🙏

josepot:
> I'm going to suspect [paritytech/polkadot-sdk#7866](https://github.com/paritytech/polkadot-sdk/pull/7866)
> 
> cc [@lexnv](https://github.com/lexnv) Could you tell us if there was some kind of intentional protocol change in the transactions? 🙏

We can confirm!

[...]

PR #2130, opened 2025-05-27
The handshake of the transactions protocol is now the Role

tomaka:
cc https://github.com/paritytech/litep2p/pull/395

Close https://github.com/smol-dot/smoldot/issues/2128


josepot:
> cc @josepot Unfortunately I don't own any KSM anymore (and for the sake of tax reporting I'd prefer to not buy some again just for testing) so I didn't test whether this actually fixes #2128.

We did test it and it works like a charm. Thanks @tomaka !

Issue #2137, opened 2025-06-06
Parse error on valid JSON-RPC messages


josepot:
According to the [JSON-RPC spec](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_v1_follow.html#possible-errors), the following request:

```json
{"jsonrpc":"2.0","id":"1-1","method":"chainHead_v1_follow","params":[]}
```

should produce the following reply:

```json
{"jsonrpc":"2.0","id":"1-1","error":{"code":-32602,"message":"Invalid parameters"}
```

However, smoldot replies with the following message:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}
```

which can't be traced back to the original request b/c it uses `null` as the id of the response.

Similarly, when sending a message that uses a non-supported method such as:

```json
{"jsonrpc":"2.0","id":"2-8","method":"archive_v1_storage","params":["0x750d7c25fd87702beabd86afd4cf1cc2a8cdcab7fae89b189900b4fcf894e0e7",[{"key":"0x3a636f6465","type":"hash"}],null]}
```

I would expect a reply like:

```json
{"jsonrpc":"2.0","id":"2-8","error":{"code":-32601,"message":"Method not found"}
```

However, we also get the exact same "Parse error" with a `null` id:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}
```

This can be easily reproduced:

```ts
import { start } from "polkadot-api/smoldot"
import { chainSpec } from "polkadot-api/chains/westend2"

const smoldot = start({
  maxLogLevel: 8,
})

const chain = await smoldot.addChain({ chainSpec })

chain.sendJsonRpc(
  `{"jsonrpc":"2.0","id":"1-1","method":"chainHead_v1_follow","params":[]}`,
)

console.log(await chain.nextJsonRpcResponse())
chain.remove()
await smoldot.terminate()
```

which produces the following logs:

```
[smoldot] Smoldot v2.0.35
[smoldot] Chain initialization complete for westend2. Name: "Westend". Genesis hash: 0xe143…423e. Chain specification starting at: 0x594d…d13e (#26215112)
[json-rpc-westend2] json-rpc-request-queued; request={"jsonrpc":"2.0","id":"1-1","method":"chainHead_v1_follow","params":[]}
[runtime-westend2] start; genesis_block_hash=0xe143…423e
[...]

PR #2138, opened 2025-06-06
Fix JSON-RPC request parsing errors being ignored

tomaka:
Close https://github.com/smol-dot/smoldot/issues/2137

Work time: 45mn

### DIFF
--- a/wasm-node/javascript/test/misc.mjs
+++ b/wasm-node/javascript/test/misc.mjs
@@ -40,6 +40,25 @@ test('malformed JSON-RPC request generates an error', async t => {
     .then(() => client.terminate());
 });
 
+test('invalid JSON-RPC method generates an error', async t => {
+  const client = start({ logCallback: () => { } });
+  await client
+    .addChain({ chainSpec: westendSpec })
+    .then((chain) => {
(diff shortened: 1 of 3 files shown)


wasm-node/javascript/test/misc.mjs
```diff
@@ -40,14 +40,33 @@ test('malformed JSON-RPC request generates an error', async t => {
...
+  await client
+    .addChain({ chainSpec: westendSpec })
+    .then((chain) => {
+      chain.sendJsonRpc('{"jsonrpc":"2.0","id":1,"method":"this method doesnt exist","params":[]}');
+      return chain;
+    })
+    .then(async (chain) => {
+      const response = await chain.nextJsonRpcResponse();
+      const parsed = JSON.parse(response);
+      if (parsed.id === null && parsed.error)
```
carlosala, line 54: Actually, this shouldn't be `null`, right? It should be the `id` sent.

tomaka: Indeed. The tests take a very long time to run for me locally (several minutes), and I ran out of patience before it hit the new test.

carlosala: I understand the feeling, our zombienet tests have the same problem :)

carlosala:
Thanks for the prompt fix Pierre <3

PR #2166, opened 2025-08-29
Update slab to v0.4.11

tomaka:
Previous version of slab contains a "vulnerability".
This doesn't concern us, but it makes the CI of https://github.com/smol-dot/smoldot/pull/2165 happy


tomaka:
I'm force-merging, as the only failing CI check is the audit due to https://github.com/smol-dot/smoldot/pull/2167

PR #2167, opened 2025-08-29
Try update wasmtime to v36

tomaka:
Work time: 10mn


tomaka:
I'm force-merging, as the only failing CI check is the audit due to https://github.com/smol-dot/smoldot/pull/2166

Issue #2089, opened 2025-02-27
Repeated `operationInaccessible` on call `TaggedTransactionQueue_validate_transaction` for `System.apply_authorized_upgrade`


voliva:
Before broadcasting transactions, polkadot-api validates them with a runtime call to TaggedTransactionQueue_validate_transaction.

We have this issue in papi-console https://github.com/polkadot-api/papi-console/issues/35 where they wanted to try sending a `System.apply_authorized_upgrade` (unsigned tx) for polkadot 1.4, but the transaction doesn't get broadcasted.

Looking at the JSON-RPC logs, it seems that the call gets stuck in a loop of `operationInaccessible`, and retrying it after a few ms.

Looking at the logs I can see that smoldot internally retries a few times until it fails as well:

```
smoldot[json-rpc-polkadot(4)] json-rpc-request-queued; request={"jsonrpc":"2.0","id":"2-46","method":"chainHead_v1_call","params":["GtJnfhu95ow96MjjAoA5bWczUeW4K5Tb3U1BKJ68qD8c","0xe895f3db599e2dc8d7209b48a864822c12310e3bdae7060e62131af9803c6888","TaggedTransactionQueue_validate_transaction","0x02eee77100040002d…
smoldot[runtime-polkadot(5)] foreground-pin-pinned-block-runtime; subscription_id=4, block_hash=0xe895…6888
smoldot[runtime-polkadot(4)] foreground-runtime-call-start; block_hash=0xe895…6888, block_number=24913412, block_state_trie_root_hash=0x5f1e…1fb4, function_name=TaggedTransactionQueue_validate_transaction, required_api_version=None, parameters_vectored=0x02ee…6888, total_attempts=3, timeout_per_request=20s
smoldot[runtime-polkadot(5)] foreground-runtime-call-request-start; block_hash=0xe895…6888, function_name=TaggedTransactionQueue_validate_transaction, parameters_vectored=0x02ee…6888, call_proof_target=12D3KooWS9ZcvRxyzrSf6p63QfTCWs12nLoNKhGux865crgxVA4H
smoldot[network(4)] call-proof-request-started; chain=polkadot, target=12D3KooWS9ZcvRxyzrSf6p63QfTCWs12nLoNKhGux865crgxVA4H, block_hash=0xe895…6888, function=TaggedTransactionQueue_validate_transaction
[...]

Issue #1526, opened 2023-12-28
`transaction_unstable_submitAndWatch` fails with large transactions


josepot:
I tried creating an extrinsic against Westend consisting in a `Utility.batch_all` call with 1200 `System.remark_with_event` calls of 1kb each, and I consistently receive the following error:

> Error when retrieving the call proof: Call proof query errors:\n- Timeout

[The logs.](https://github.com/smol-dot/smoldot/files/13785031/smoldot-logs.txt). Also, the code that I ran to get this logs and to test this can be found in the [following repo](https://github.com/josepot/smoldot-large-tx-test).

tomaka:
Looking at the logs, when we start asking for the call proof, there's a few packet exchanges, and then smoldot sends several packets of 8222 bytes each, which I assume in the transaction. After these packets have been sent, no incoming data is ever read on the connection, and eventually the call times out.

The packets of 8222 bytes only total to around 150kiB of data. My assumption is that smoldot doesn't send more than that amount because it waits for a Yamux ACK packet before doing so, except that this packet never arrives.

The logs unfortunately stop 20 seconds after sending the transaction. During these 20 seconds we don't read any data back, which is very suspicious.

josepot:
I repeated the test, but this time I let smoldot run for 3 more minutes after having received the error (so that we can capture more logs). [These are the logs of that attempt](https://github.com/smol-dot/smoldot/files/13785525/smoldot-logs.txt)

tomaka:
Sending large transactions is unfortunately not something that can be easily fixed, due to the existence of a limit to the size of a gossip message.

In the second set of logs, the remote shuts down the connection when this limit is reached, which is the expected behavior.

In the first and third sets of logs, the remote instead simply stops sending anything back altogether. In the third set of logs, we see that it doesn't even try pinging us back. I'm very curious about the reason why this happens, but the logs don't really help figure it out.

bkchr:
> Sending large transactions is unfortunately not something that can be easily fixed, due to the existence of a limit to the size of a gossip message.

[...]

PR #2196, opened 2026-01-09
Support execution proofs with inputs bigger than 1MiB

bkchr:
Right now execution proofs with an input bigger than 1MiB fail to execute. The problem is that the req-resp configuration only allows up to 1MiB for the initial request. Such a big input may happens for `validate_transaction` when doing for example a runtime upgrade.

This pull request fixes this by checking if the input is too big and if this is the case, it starts an on demand execution. This means it starts executing the function and requests storage entries on demand. This increases the latency when executing functions that require a big input, but these should not be that common any way. On the positive side this makes smoldot support these big inputs without requiring any protocol changes.

Closes: https://github.com/smol-dot/smoldot/issues/1526
Closes: https://github.com/smol-dot/smoldot/issues/2089

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.25.0"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9acbfca36652500c911ddb767ed433e3ed99b032b5d935be73c6923662db1d43"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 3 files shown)


light-base/src/runtime_service.rs
```diff
@@ -660,6 +660,10 @@ pub enum RuntimeCallInaccessibleError {
     InvalidCallProof(proof_decode::Error),
     /// One or more entries are missing from the downloaded call proof.
     MissingProofEntry,
+    /// Failed to download a storage proof from the network (storage-on-demand mode).
+    StorageRequest(network_service::StorageProofRequestError),
+    /// Failed to download a child storage proof from the network (storage-on-demand mode).
+    ChildStorageRequest(network_service::ChildStorageProofRequestError),
```
tomaka:
```suggestion
    /// Failed to download a storage proof from the network.
    StorageRequest(network_service::StorageProofRequestError),
    /// Failed to download a child storage proof from the network.
    ChildStorageRequest(network_service::ChildStorageProofRequestError),
```

This punches a bit through the abstraction layer, so I don't think this needs to be specified.

tomaka (approved):
Thanks! The solution actually seems so obvious that I wonder why we didn't do this before.

I think it would have been easier to support child tries by just adding an `child_trie: Option<Vec<u8>>` field to the existing code for storage requests, but since you did the work already, it's fine as it is.

I couldn't really spot any mistake except for a few nits. I'm not a huge fan of the logic in `runtime_service.rs` (I feel like we could de-duplicate a lot of code), but it's hard to tell you how it should be done without trying myself to implement it differently, so I think it's fine as it is as well.

Did you try if it works? The logic in runtime_service is a bit hard to read, but if you tried if it worked then that's good enough for me.

Could you also add an entry at the top of `CHANGELOG.md`?

bkchr:
@tomaka applied your comments.

Issue #2158, opened 2025-08-14
Missing and delayed blocks on parachains


josepot:
This issue is a followup of [this discussion](https://github.com/smol-dot/smoldot/discussions/2156#discussioncomment-14093589).

On parachains smoldot reports new blocks with a considerable delay, and in many instances it never seems to report blocks that are not part of the current best-block chain.

I have created [a small script](https://github.com/josepot/smoldot-slow-blocks/blob/main/index.ts) to dig deeper into this.

What the script does is to subscribe to the new blocks received by both the smoldot and a centralized RPC for some blocks, and at the end it compares the blocks that smoldot did not report.

I'm attaching the logs of a particular run:

- [Smoldot low-level logs](https://github.com/user-attachments/files/21774928/SM_LOGS.txt).
- [Smoldot JSON-RPC logs](https://github.com/user-attachments/files/21774936/SM_JSON_RPC.txt) (the JSON-RPC messages sent/received between the smoldot chain and PAPI)
- [Websocket JSON-RPC logs](https://github.com/user-attachments/files/21774938/RPC_JSON_RPC.txt)

And this is the output of said run:
```
Blocks that smoldot did not report:
{
  hash: "0xed87ee644095072d48c1b9eb0b023284dc4e4fceb68fadd75e17b771ca5edaa5",
  height: 9484930,
}
{
  hash: "0x816fcd370dc0479bc5c7cccd9b4b7e7288c67e764e32bc62861da7f20fe34a18",
  height: 9484931,
}
{
  hash: "0xfe46486dc4016f090fa6bd1059d7ed85bcd52ca929314f4bc16d138bcbaadcee",
  height: 9484932,
}
{
  hash: "0xaa2ed7ad331db94cb8db95588fd1566da99d2994afe27f5bbc5ef56ca959b9ca",
  height: 9484932,
}
{
  hash: "0x25f11905d5459bc9d6b98719f12837feabe60160edd12edb7d20e3798c1de666",
  height: 9484934,
}
{
  hash: "0xf62e41976f7749f21966451d7454732e608956607f49c22fedb86db92fce2f5f",
  height: 9484936,
}
{
  hash: "0xea1cd36afacd2aeddc97f75c6f49fe5335c2c42cb9fbbf88f3660c71de6334e6",
  height: 9484936,
}
{
  hash: "0xb9cda13896522a63977509978e147abdfa316ff5c11d70425a4d709a8d51d27c",
  height: 9484937,
}
{
[...]

PR #2198, opened 2026-01-12
[WIP] Add support for Substrate-compatible parachains

tomaka:
This is a work-in-progress PR to add support for parachains that sync with the parachain network the same way as standalone/relay chains, but also get the finality information from the relay chain.

I went through the code, and contrary to what I've said in #2158 I think it's better to plug this to the standalone chain code. 

The general plan would be:

- [x] Try to sync a parachain (without finality) using the existing standalone chains code and fix the potential problems that arise.
- [x] Extract the code in `light-base/src/sync_service/parachain.rs` that does runtime request into a separate file that gives events about the relay chain, and update `parachain.rs` to use this code.
- [ ] Also use this code in `substrate_compat.rs`, adding any missing function in `lib/src/sync` such as a function to force-set the current finalized block.
- [ ] Add some way to specify whether the chain is Substrate-compatible from the public JavaScript API. It's completely unclear to me how, but in terms of efforts it should be minimal anyway.
- [ ] Consider removing the old `parachain.rs` altogether? Is there any parachain out there that doesn't use Substrate or has so heavily customized Substrate that it can't sync using smoldot?


tomaka:
One thing that's not clear to me is whether the parahead from the relay chain best block is important in some way or not.
From the perspective of the parachain, we can know the parablock that is likely to be finalized, but when it comes to smoldot we don't actually care about this information since we don't propagate it through the JSON-RPC layer.

cc @bkchr 🙏  Or is there something I'm forgetting?

tomaka:
The new `paraheads.rs` file is basically the same as the existing `parachain.rs` but without any networking. It simply reports the new and finalized paraheads by following the relay chain and does nothing else.
Then `parachain.rs` combines this with the networking. Normally no behavior has changed so far combined to the main branch.

The next step will be to make the "substrate-compatible" sync service use this paraheads service when relevant.
I am on vacation for the next three weeks, but I'll go back to it afterwards.

bkchr:
@tomaka I wish you nice vacations.

tomaka:
I'm now back from vacation. However, the plan was to fund my work through PAPI. Given that [PAPI no longer has funding](https://forum.polkadot.network/t/polkadot-api-project-update/16951), I'm also stopping work on this PR.

bkchr:
Hey @tomaka,
would you still be available to at least merge this: https://github.com/tomaka/smoldot/pull/1 into your branch and then into master?

I thought about removing `parachain`, but found it logically better separated to keep it and just wrap the other stuff. I tested everything and it works flawlessly as far as I can tell.

tomaka:
I'm still waiting for an answer from Parity as to whether I'll get paid to work on smoldot, but given that the paritytech/smoldot repository is seeing activity again I'm assuming that the answer is no.

PR #2214, opened 2026-02-03
Fix Notifications close desired timeout handling

dmitry-markin:
Force-close inbound Notifications substream after (not before) the close desired timeout is reached.


tomaka (approved):
Thank you!

PR #2248, opened 2026-05-02
runtime_service: don't panic if to_background_tx fails to upgrade after chain removal

### LINKED ISSUE #3238: Smoldot panic after initial parachain sync if the chain object was removed
After updating PAPI to use 3.1.1, I've seen in the dapps I've upgraded that smoldot panics after syncing. I'm still not sure why, but so far it only happens in browser (i.e. the PAPI CLI doesn't have any issue connecting to a parachain)

Smoldot will take some time to initialize, and then panic with:

```
Smoldot has panicked while executing task `runtime-asset-hub-polkadot`. This is a bug in smoldot. Please open an issue at https://github.com/smol-dot/smoldot/issues with the following message:
panicked at /__w/smoldot/smoldot/light-base/src/runtime_service.rs:1689:78:
called `Option::unwrap()` on a `None` value
```

Logs 3.1.1 (KO)
[smoldot_3.1.1.txt](https://github.com/user-attachments/files/27235021/smoldot_3.1.1.txt)

Logs 3.1.0 (OK)
[smoldot_3.1.0.txt](https://github.com/user-attachments/files/27235020/smoldot_3.1.0.txt)

SAY-5:
Fixes paritytech/smoldot#3238 (the same code path lives here at `light-base/src/runtime_service.rs:1689`).

When a chain object is removed concurrently with an in-flight `SubscribeAll` response, `background.to_background_tx.upgrade()` returns `None` and the `unwrap()` panics the runtime task, this is the panic users started hitting after upgrading to 3.1.1 in browser dapps. The receiver is going away anyway in that case; skip the response with `continue`.


voliva (commented):
I'm not familiarised enough with the codebase, but I'm wondering... the `.unwrap()` has been there for quite a long time, and it hasn’t caused issues until 3.1.1.

To me, it looks like it was written based on an assumption that may no longer hold true, and that could matter. Simply turning it into a noop when `background_tx.upgrade()` returns `None` might have other implications as well.

My question is, what’s actually causing this channel to close instead of cleanly shutting down whatever process it was handling?

